PR #15, opened 2019-04-04
Dev root6

knakano0524:
This version is aimed at making e1039-core compatible with ROOT 6 as well as ROOT 5.

The compilation with ROOT 6 creates a "*_rdict.pcm" file per "*LinkDef.h", which should be installed into "lib/".  In this version, almost all CMakeLists.txt files have additional lines to install the pcm file when ROOT > 5.

Two methods for the pcm installation exist for now.  One is more strict, which makes the list of "*_rdict.pcm" files from the list of "*LinkDef.h" files under one-to-one correspondence. But sometimes, in "database/pdbcal/base" for example, a LinkDef.h file is described but not created. Thus a simpler method was used later, which just copies all existing pcm files in the build directory. We might choose one of them in the future.

The key step next to this update is to merge the "dev_seaquestdaq01" branch to the master branch (otherwise the online and offline parts will separate largely).  Then we can develop the SQ interface objects etc.


HaiwangYu:
Thanks a lot for this ROOT6 compatible update. It is a lot of works. I am testing it on both seaquestdaq01, ROOT6 and seaquestgpvm01, ROOT5.

 - On seaquestgpvm01 with ROOT5:
The e1039-core libraries are compiled successfully 
The "Fun4TrkDev" tracking macro (contains all modules used in reconstruction chain) was also run successfully

 - On seaquestdaq01 with ROOT6:
I used your dependencies by `source /opt/e1039-share/this-share.sh`
Then the e1039-core libraries are compiled successfully
However, seem some linking error happened when I tried to run the tracking macro.
It seems the 'libG4OpenGL.so' is missing.

I will keep working on it tomorrow.
PS: I uploaded the macros [here](https://github.com/E1039-Collaboration/e1039-analysis/tree/master/PrgTrkDev)
My working dir on seaquestdaq01 is here:
/home/yuhw/e1039-analysis/PrgTrkDev

knakano0524:
Thank you for the tests, including running the macro (which I haven't tried).  About 'libG4OpenGL.so', this file itself exists in "/opt/e1039-share/inst/lib64".  Most likely the cause is that this "lib64" directory is not included in LD_LIBRARY_PATH.  Can you change the corresponding line in setup-e1039-core.sh as follows and try again?
----
    export LD_LIBRARY_PATH=$MY_INSTALL/lib:$OFFLINE_MAIN/lib:$OFFLINE_MAIN/lib64:$LD_LIBRARY_PATH
----

HaiwangYu:
Thanks for updating the settings on seaquestdaq01. After the updating. The dependent libraries could be properly resolved now. And the `.so`libraries could be loaded in Fun4All macro as well. However some modules like the `JobOptSvc` threw out exception `std::logic_error` when trying to use the `init` function. I think some modules need modification to be compatible with ROOT6. 

I suggest we merge this pull request now, since this PR does kept ROOT5 compatibility. And we can work on the ROOT6 gradually in the future. @knakano0524 What do you think?

knakano0524:
I agree to merge this right now.  Note that I cleaned up setup-e1039-core.sh slightly.  I remember to have seen the std::logic_error error.  I expect we can fix it in the near future.

HaiwangYu:
Great. I will merge it then.

PR #22, opened 2019-04-23
Implementations of calibration parameters and calibrator classes.

knakano0524:
I added the CalibParam* classes which handle the calibration parameters (XT curve and in-time window) in the same way as the channel mapping (namely the run range and the DB access handled in ChanMapTaiwan).  I rearranged many classes, functions and variables so that the names and inheritance relations of all classes are more clear.

I then added two calibrator classes (CalibInTime & CalibXT), which read the parameters from CalibParamInTimeTaiwan etc. and accordingly set the in-time and drift-distance fields of SQHit.  They are enabled in Fun4MainDaq.C.

I have confirmed that this version runs fine using Fun4MainDaq.C, run 28700, ROOT 6 and seaquestdaq01.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -23,27 +23,27 @@ else
     framework/fun4all
     interface_main
     online/chan_map
-#    online/onlmonserver
-#    online/decoder_maindaq
-#    packages/Half
-#    packages/vararray
-#    database/pdbcal/base
-#    database/PHParameter
-#    packages/PHGeometry
-#    packages/PHField
-#    generators/phhepmc
-#    generators/PHPythia8
-#    simulation/g4decayer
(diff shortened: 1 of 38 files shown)


HaiwangYu:
The path `/data2` is not defined on seaquestgpvm01.
But the `/seaquest` is visible (disk is actually mounted)
Can we use `/seaquest`? or can we define `/data2` on the gpvm?

knakano0524:
Thank you for the good catch.  Let's use /seaquest as it is visible on seaquestdaq01 also.  I will modify this branch and make a pull request again.  I hope E1039 won't have this kind of troubles, i.e. differences between servers.

PR #26, opened 2019-05-03
Onlmon client

knakano0524:
I added two online-monitor classes (OnlMonSpill and OnlMonCham).  OnlMonCham is a typical example of subsystem basic monitors, which monitors the distributions of element IDs and tdcTimes.
The viewer macro (OnlMon4MainDaq.C) was also updated so as to select the monitor to be drawn, as seen in the attached screenshot.

onlmon_example_20190503

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -25,25 +25,25 @@ else
     online/chan_map
     online/onlmonserver
     online/decoder_maindaq
-    packages/Half
-    packages/vararray
-    database/pdbcal/base
-    database/PHParameter
-    packages/PHGeometry
-    packages/PHField
-    generators/phhepmc
-    generators/PHPythia8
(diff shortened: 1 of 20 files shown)


HaiwangYu:
This is really great!
I compiled the code and run it on seaquestgpvm01. Got same results:
image

Just I think it might be useful to keep the macros ROOT5 compatible for now. I made a pull request as an example.

Also seem the "D0" plot is not implemented. If I click the "D0" button, both server and client would abort. We may need some better exception handling mechanism in the future. But for now it is already amazing!

knakano0524:
It is correct that the D0 plot and any exception handling are not implemented for now.  I will work on them next, together with more subsystem monitors.

PR #29, opened 2019-05-10
New SubsysReco modules to upload run & spill info to MySQL DB

knakano0524:
DbUpRun and DbUpSpill are new SubsysReco modules to upload run & spill info to MySQL DB (namely "run" and "spill" tables in "user_e1039_maindaq" schema on e906-db1).

The SQ interface was updated to include the new run & spill info.  This time I didn't make v2 but modified v1.  Also the OnlMon module for MainDAQ (i.e. OnlMonMainDaq) was updated to monitor the new info.

I removed some obsolete variables in DecoParam and moved some to DecoData.


knakano0524:
I note three things that are not related to the main aim of this branch.

1. I have fixed the bug that Haiwang found about handling the width of the in-time window.  It is just two lines changed in CalibInTime.cc.

2. "generators/phhepmc/CMakeLists.txt" references "E1039_SHARE".  This variable doesn't exist in my environment.  Mine defines and uses "DIR_E1039_SHARE" in /opt/e1039-share/this-share.sh.  The difference is not only the variable name but also its pointing level, namely "/opt/e1039-share/inst" by "E1039_SHARE" versus "/opt/e1039-share" by "DIR_E1039_SHARE".  We should choose one of them.

3. I cannot compile "packages/db2g4/DetectorConstructionWrapper.cxx" because "FMAGSTR" no longer exists in "GlobalConsts.h".  I guess the latest version of the cxx file(s) hasn't been committed.  Or is my environment not properly set up?

HaiwangYu:
Thanks for this update.

I tried to build this branch on the gpvm with ROOT5.
Seem CINT doesn't like this function:
https://github.com/E1039-Collaboration/e1039-core/blob/db_uploader/packages/db_svc/DbSvc.h#L35
``` c++
void CreateTable(const std::string name, const int n_var, const char* list_var[][3]);
```
The error msg:
```bash
[ 66%] [100%] Building CXX object CMakeFiles/db_svc.dir/DbSvc.cc.o
Building CXX object CMakeFiles/db_svc.dir/DbSvc_Dict.cc.o
/e906/app/users/yuhw/e1039-core-build/build/packages/db_svc/DbSvc_Dict.cc: In function ‘int G__DbSvc_Dict_170_0_10(G__value*, const char*, G__param*, int)’:
/e906/app/users/yuhw/e1039-core-build/build/packages/db_svc/DbSvc_Dict.cc:313:12: error: no matching function for call to ‘DbSvc::CreateTable(const string&, int, char* [1][3])’
 , G__Ap2->a);
            ^

```

I can temporarily fix this by hiding this function from CINT:
```c++
#ifndef __CINT__
void CreateTable(const std::string name, const int n_var, const char* list_var[][3]);
#endif
```
Just I am not sure if this function need to be called in an interactive ROOT session?

After this tmp fix, I compiled this branch and tested it via the [Fun4CODA.C](https://github.com/E1039-Collaboration/e1039-analysis/blob/master/CODAChainDev/Fun4CODA.C)
The results looks good.

HaiwangYu:
For the second post:

1, Thanks for the quick fix.

2, Yes, the `$E1039_SHARE ` is supposed to track the `/opt/e1039-share/inst` equivalent. As we discussed in last meeting, I was thinking to have some structure like this:
I think we could continue our discussion tonight on this.
image


3, `packages/db2g4` is obsolete, and should be removed from the compiling list. I should did that. Sorry about this.

HaiwangYu:
@knakano0524 The `dptrigger` does not rely on g4main.
https://github.com/HaiwangYu/e1039-core/blob/master/packages/dptrigger/CMakeLists.txt#L35
In principle, it could be run on real data to cross-check the hardware trigger implementation/geometry alignment.
But I think the compiling sequence is arbitrary in this case as the `dptrigger` is more like an analysis module with no other modules using it.

knakano0524:
About the dptrigger, I did fail at compiling it when using the original build order in build.sh.  It was because DPTriggerAnalyzer.cxx includes g4main/PHG4TruthInfoContainer.h in line 22 and thus the g4main package should be installed beforehand;
https://github.com/HaiwangYu/e1039-core/blob/master/packages/dptrigger/DPTriggerAnalyzer.cxx#L22

I expect anyone can reproduce this compilation error when starting from scratch.

HaiwangYu:
Interesting. Sorry I must have overlooked this. It should not include those headers.
I will fix it in the future.
Thanks for catching this.

PR #34, opened 2019-05-27
Interface for CalibDB config.

HaiwangYu:
Add interface for CalibDB config.
This is needed for the default CalibDB config is not accessible from the grid.

This style is let each module configure its own configurations instead of using a JobOpt card to configure all modules. This is the prefered style in Fun4All.

Corresponding macros: [here](https://github.com/E1039-Collaboration/e1039-analysis/tree/master/CODAChainDev).


knakano0524:
It is indeed necessary that the config file (my.cnf) is accessible from the grid.  For this purpose, isn't it easier for us to add more locations of the config file to DbSvc::DbSvc()?  For example;
----
if (usr_id == Guest) {
  m_my_cnf = "/seaquest/analysis/kenichi/e1039/my.cnf";
  if (! gSystem->AccessPathName(m_my_cnf)) m_my_cnf = "/e906/app/software/osg/users/yuhw/e1039/resource/db_conf/my.cnf";
  if (! gSystem->AccessPathName(m_my_cnf)) cerr << "ERROR...";
} else { // == Prod
  // do similarly
}
----

The key info of the config file is just the password, which is unique per server and user.  So it should be enough for each module to choose only server and user (via DbSvc::DbSvc(svr_id, usr_id)), and then the location of the config file is managed by DbSvc solely.

If the location of the config file is much adjustable, I'm afraid that many users create many config files, which contain the password.

HaiwangYu:
Thanks @knakano0524 for the suggestions.
I agree with you.
I proposed an alternative way to do this in #35 
Could you have a look?
Thanks.

PR #36, opened 2019-05-29
Updates for E1039 Main DAQ

knakano0524:
This version is to prepare for the data from the new E1039 Main DAQ.  I have confirmed that this version can be compiled fine on seaquestdaq01 and seaquestgpvm02.

The decoding algorithm was tested and updated, using run 46 from the 1st-stage DAQ. Several parts, such as the handling of standard-physics events, have to be yet updated as the DAQ configuration is updated or finalized.

I added a function (via Daemon4MainDaq.C) to auto-start a decoding process when a new Coda file appears.  It should work fine, but I ought to test it once the location of Coda files is adjusted (note that the current location is not accessible over NFS) and a bunch of Coda files are recorded.


HaiwangYu:
Sorry for this late response. I spent some time setting up the new **spinquestgpvm01.fnal.gov** with SL7.
Now the **spinquestgpvm** runs ROOT6 by default.

I tested this pull request on **spinquestgpvm** with your e1039-analysis/AnaRealDst.
It worked well.

However the coda chain ([e1039-analysis/CODAChainDev](https://github.com/E1039-Collaboration/e1039-analysis/tree/master/CODAChainDev)) seems to have some problem:
```
Registering OutputManager DSTOUT
executing run for input master DefaultSyncManager
  ProcessCodaPrestart 24173 1484355778 20
Read channel map from user_e1039_chan_map_taiwan.param_e906run28740.
Read channel map from user_e1039_chan_map_v1495.param_e906run28740.
Read channel map from user_e1039_chan_map_scaler.param_e906run28740.
  run desc: 5304 chars.
feePrescale: codaEventID = 21
!!ERROR!!  idx != idx_end in StdFeeQIE().
274 != 49605
```

Probably I did something wrong.
I can do some more tests tomorrow and give you some more feedbacks.

knakano0524:
Thank you for the check.  I reproduced this problem in my environment on seaquestdaq01, using run 24173 and around it.  Run 28700 is fine.  I have disabled a part of codes to avoid this problem.  The part is not necessary for the E906 FY2017 runs, and also E1039 runs at present.

HaiwangYu:
Thanks! I tested with run 28692. It did work on it.
I will merge this PR if you agree.

knakano0524:
I agree to merge.  Please go ahead.

PR #38, opened 2019-06-10
Small update on OnlMonServer for test of network communication.

knakano0524:
I included in OnlMonServer the functions that were defined in "HistoServer.cc" (which was not a class but a set of global functions).  I confirmed that the OnlMon server and client runs fine on seaquestdaq01.

I initially intended to test the OnlMon program over network, namely the communication between seaquestdaq01 (=server) and spinquestana1 (=client).  But I haven't tried it because the network is not ready yet.  I want to merge this branch now in order to suspend the test.


HaiwangYu:
Compiled on spinquestgpvm with ROOT6. Tested with Sim chain and CODA chain.

PR #43, opened 2019-06-17
Init PatternDBGen; Sim. compatible with run6 config

HaiwangYu:
Init PatternDBGen for dedicated DSearch tracking database simulation.

Updated GeomSvc and related modules to make the simulation compatible with run6 configuration (`$E1039_RESOURCE/alignment/run6`) and @liuk 's geometry schema `user_liuk_geometry_G9_run5_2`

**Seems this schema used detector id 7-12 for the DP trigger planes.
So DC1 shifted to use 1-6. This may won't be the case in the future.**

[e1039/analysis/CODAChainDev](https://github.com/E1039-Collaboration/e1039-analysis/tree/master/CODAChainDev) and [e1039/analysis/SimChainDev](https://github.com/E1039-Collaboration/e1039-analysis/tree/master/SimChainDev) updated accordingly.

### DIFF
--- a/packages/ktracker/PatternDBGen.cxx
+++ b/packages/ktracker/PatternDBGen.cxx
@@ -1,0 +1,372 @@
+/**
+ * \class PatternDBGen
+ * \brief General purposed evaluation module
+ * \author Haiwang Yu, yuhw@nmsu.edu
+ *
+ * Created: 08-27-2018
+ */
+
+
+#include "PatternDBGen.h"
+
+#include <interface_main/SQHit.h>
+#include <interface_main/SQHit_v1.h>
+#include <interface_main/SQMCHit_v1.h>
+#include <interface_main/SQHitMap_v1.h>
+#include <interface_main/SQHitVector_v1.h>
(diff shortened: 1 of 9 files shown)


packages/ktracker/PatternDBGen.cxx
```diff
@@ -0,0 +1,375 @@
...
+#include <interface_main/SQHit_v1.h>
+#include <interface_main/SQMCHit_v1.h>
+#include <interface_main/SQHitMap_v1.h>
+#include <interface_main/SQHitVector_v1.h>
+#include <interface_main/SQEvent_v1.h>
+#include <interface_main/SQRun_v1.h>
+#include <interface_main/SQSpill_v1.h>
+#include <interface_main/SQSpillMap_v1.h>
+
+#include <ktracker/SRecEvent.h>
```
knakano0524: This line should be changed to #include "SRecEvent.h".  Otherwise a build from scratch fails.

HaiwangYu: Thanks! I will change it in the next PR.

PR #46, opened 2019-06-26
Updates for better stability of online decoding.

knakano0524:
I made various updates to improve the stability of the online decoding.  I changed also some default settings about data & parameter files to match with the real situation.

This version is not yet complete, as for example the auto-cycle mode of the online monitor always crashes.  But the problem is isolated in the online part.


HaiwangYu:
Thanks for the update. I compiled the code on `spinquestgpvm01` from scratch and tested with the Sim chain and CODA chain.

PR #48, opened 2019-06-29
Geometry info including DP detector via DbSvc.

knakano0524:
The aim of this update is to include the DP detector in the E1039 software framework.

I added a DbSvc-based manager class (GeomParamPlane) and created a new geometry plane info on MySQL DB, namely "user_e1039_geom_plane.param_G9_run5_2" on e906-db1.fnal.gov.  I added a switch function (UseDbSvc()) to GeomSvc so that we can select using the new geometry info.

In order to use the DbSvc-based info, one should call "GeomSvc::UseDbSvc(true)" at the beginning of ROOT macro (to be pricise, before "GeomSvc::instance()" is called first).

This version worked fine when I decoded run 246.  But this test is not complete since it doesn't use the plane geometry but just the detector ID.  It is very helpful if Haiwang could compare tracking results (using simulated events)  between the original and DbSvc-based geometry info.


HaiwangYu:
Thanks a lot for making this systematic change to the geometry related utilities. I am working on testing it.

HaiwangYu:
Thanks a lot for this systematic change to the geometry related utilities. I updated the macros in the [e1039-analysis bfc8732](https://github.com/E1039-Collaboration/e1039-analysis/commit/bfc87327a2cb2f4178d938005d94bf9de41790f5) to match with this pull request.
**Please update your macro for this PR, refer to the link above**

Key changes needed for the macro:
- run7_xxx.opts added to use Kenichi's geom schema, `user_e1039_geom_plane`, containing the Dark Photon detectors
- The structure in `user_e1039_geom_plane` is different from previous geometry schema like `user_liuk_geometry_G9_run5_2` One needs to set `GeomSvc::UseDbSvc(true)` to read that into the GeomSvc

**DP detector G4Hits are generated as below**
image



**TrkEval and the event display has not been updated to include the DP detectors**, I am working on it.

## initial tests for this PR:

### geometry display:
The 8 DP detectors are located at 798cm and 1471cm:
image

### Sim chain test using DY compared with #44 
Consistent within uncertainty
image

### CODA chain test using E906 data, run28692
Exact same results
image

PR #55, opened 2019-09-05
Updates on the online code in Aug 2019

knakano0524:
The online code (decoder & online monitor) was updated as follows in Aug 2019.  It is running fine on seaquestdaq01 as the online-version library (/data2/e1039/core-online).  It is not urgent but now I like to merge the updates.
 
1. Daemon4MainDaq.C, which starts a decoding process per run, now checks and waits until the Coda file directory is mounted.
2. The online monitor for Main DAQ (OnlMonMainDaq) shows the number of Taiwan TDC boards per event.  A deviation of this number tells us a problem on board(s).
3. The OnlMon server (OnlMonServer) rejects all client connections that come from WAN.  It has improved the security and the stability (where the server process died due to port scans multiple times in the past), although such connections should be blocked first by the firewall.

### DIFF
--- a/online/decoder_maindaq/CodaInputManager.cc
+++ b/online/decoder_maindaq/CodaInputManager.cc
@@ -42,7 +42,7 @@ int CodaInputManager::OpenFile(const std::string fname, const int file_size_min,
     sleep (sec_wait);
   }
   if (! size_ok) {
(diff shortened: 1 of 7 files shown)


knakano0524:
Thank you for taking care of this request.  Just for our information, I have confirmed that the updated master branch was successfully built and installed to /data2/e1039/core.

PR #57, opened 2019-10-08
Modification to store unmerged H4 hits (H4Tu, H4Td, etc.) in DST.

knakano0524:
This update is to store in DST the unmerged H4 hits (namely H4Tu, H4Td, H4Y1Ll, H4Y1Lr, etc.).  I hope it will be pulled in a few days so that the hodoscope expert (i.e. Forhad and Steve) can use it in the HV tuning.

In the past scheme, using H4Tu and H4Td as an example,
* GeomSvc.cc defines only one ID for H4T.
* The channel mapper (ChanMapTaiwan) assigns this H4T ID to both H4Tu and H4Td.
* The decoder uses it in converting (roc, board, chan) to (det, ele).
* The DST file contains H4T hits, with no difference between H4Tu and H4Td.

In the new scheme
* H4Tu, H4Td and H4T are given each unique ID in GeomSvc.cc.
* The channel mapper takes care of H4Tu and H4Td, where H4T doesn't appear because of no hardware (i.e. roc, board, chan) for H4T.
* The decoder hasn't changed.
* The DST file contains H4Tu and H4Td hits separately (and no H4T hit).
* A new calibrator ("CalibMergeH4"), when enabled, adds H4T hits to the hit list in DST.

CalibMergeH4 by default merges all hits in the "or" mode as the past scheme does.  I have confirmed that this "or" mode results in the same hit distribution as the past scheme, by comparing the OnlMon hit distribution of run 843.

CalibMergeH4 by default keeps unmerged hits (like H4Tu) after adding merged hits (like H4T) to DST.  I suppose this behavior doesn't disturb any analysis module that reads DST, particularly the tracking, as long as it extracts hits using a set of detector IDs that needs.  This had better be confirmed by Kun et al.

OnlMonH4 is a new OnlMon client that analyzes the unmerged H4 hits.  Attached are its output for example.  The 1D histograms are the element and time distributions of all hits.  The 2D histograms show the hit correlations in element and time when each detector (=PMT) has just one hit.

When I was testing this new scheme, I made the following changes as well for convenience.
1. Now the locations of the decoder outputs can be changed at once via "UtilOnline::UseOutputLocationForDevel()".  A future test will never overwrite the standard outputs.
    - DST file in `/data2/e1039/dst-devel` instead of `.../dst`
    - OnlMon output in `/data2/e1039/onlmon/plots-devel` instead of `.../plots`
    - DB output in `user_e1039_maindaq_devel` instead of `user_e1039_maindaq`
1. The following macros were modified or added.  They can handle the parameters to be uploaded to DB in a more convenient way.  I made a new channel mapping that includes the unmerged detector IDs,  as explained in `/data2/e1039/resource/chan_map/taiwan/2019100701/00readme.txt`.  It is not yet the default in the online decoder.
    - UploadChanMap.C ... handle the channel mapping.
    - UploadCalibParam.C ... handle the calibration parameters (like in-time and XT).
    - UploadGeomParam.C ... handle the geometry plane parameter.

OnlMonH4T_can0
OnlMonH4T_can1


knakano0524:
To test this update, I would ask Abi to generate a small set of simulated events and analyze it, so as to confirm that the updated version is good for the simulation also.  I understand that the simulation macro runs fine without any modification, because the simulation directly generates "merged" hits (like H4T, not H4Tu).

PR #58, opened 2019-10-08
Updates for better stability of MainDAQ decoder.

knakano0524:
A big issue about the MainDAQ decoder is that a decoder process sometimes (once per day) gets stuck and never ends.  I made several small updates, and here I request a pull because this new version is running without problem in the past few days.

I would expect this version doesn't conflict with PL 57.  But if it is found conflicting after PL 57 is merged, please just let me know it so that I resolve it.

I think the last update did fix the issue.  The Mutex variable in `OnlMonServer` had not been initialized well (although it was how the original example did), which sometimes caused a never-ending wait when `OnlMonClient::process_event()` called `pthread_mutex_lock()`.

Below are the other updates included in this request, which are likely not related to the issue but are helpful anyway.
1.  `build.sh` now copies all macro files to the installation directory.
1. A wrong handling of coda ID in the online decoding mode has been fixed in `CodaInputManager.cc`.
1. Now a temporary spill ID is given in `MainDaqParser.cc` when no spill-ID info is provided by the spill counter nor the slow control.  Should be helpful during the cosmic-ray commissioning.
1. Old log file written by the decoder is no longer overwritten but renamed by `Daemon4MainDaq.C` before a new log file is opened.
1.  Now `setup-install.sh` has to be executed, not sourced.  This usage should be more natural with respect to its present behavior, namely it does not set any shell variable but creates setup scripts.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -75,9 +75,12 @@ do
 done
 
 
-( ## Install all macro files.
+(
+    echo "================================================================"
+    echo "Install all macros to $install/macros/."
     cd $src
     find . -type d -regex '.*/macros*' | while read DIR_SRC ; do
+	echo "  $DIR_SRC"
 	DIR_DEST=$install/macros/$(dirname $DIR_SRC)
 	mkdir -p $DIR_DEST
 	cp -p $DIR_SRC/* $DIR_DEST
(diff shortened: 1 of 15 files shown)


knakano0524:
To test this update, I would ask Abi to just compile it on gpvm01.  I think he need not run it because I have already done that and it's hard for others to run it (using the real data).

PR #61, opened 2019-10-22
Fix the insensitive volume geometry

liuk:
This pull request address 3 issues in the insensitive volume (F/KMAG, instrumentation, target, shielding, collimator).

1. The Iron in the FMag is defined in a weird way so that the material cannot be recognized when converted from Geant4 to GDML. Thus all the reconstruction is assuming an empty FMag. This is fixed by creating materials using standard NIST material database;
2. In the construction script some Geant4 units are missing. One obvious example is the hole on the front face of FMag is somehow moved to the center. It has been fixed in this PR;
3. There are a lot of legacy objects in the beamline (like helium bag) which will not show up in the E1039 beamline. I started with a new geometry schema. The sql script to create this schema is included in $E1039_CORE/macros

This is the FMAG geometry w/o this PR.
FMAG geometry w/o this PR

This is the FMAG geometry w/ this PR.
FMAG geometry w/ this PR


knakano0524:
The updates should be mandatory for the simulation.  Thank you for making them.  The cc/h files changed here seem fine, although I have/will not run it for test since I don't have a suitable environment.

I would like to discuss how we will manage the geometry info, i.e. the usage of `macros/geom_e1039_run1_v1.sql`. Will we create an SQL script per configuration?  It's hard for me to find and/or change a parameter of interest out of this long script.

I think we better use the existing system in e1039-core.  It reads from TSV files, although it handles only the `Planes` info for now.  The parameter set is kept in `/data2/e1039/resource/geom/plane/G9_run5_2/param.tsv` and uploaded by `e1039-core/packages/geom_svc/macros/UploadGeomParam.C`.  One key function is that one configuration is auto-selected according to run number.

abinashpun:
As far as I understand, my next goal (after I finish the physics generator) is to create a class for insensitive detectors (just like GeomSvc for sensitive). After that, we can configure the geometry via set-up macro just as for the sensitive detector case. 

Should I go ahead and merge the branch for now?

knakano0524:
To Abi,
When we use the existing system in e1039-core, I will create an interface class per parameter set like `GeomParamPlane`, which reads and writes parameters from/to TSV and DB.  `PHG4E1039InsensDetector` can directly use it.

Or, if we expect that the insensitive volume geometry will be hardly updated, we can separate it from the often-updated geometry (like `Planes`) and manage it differently.

I myself am fine with merging this branch if you and Kun would agree that we won't use SQL scripts to manage at least the often-updated geometry.

liuk:
To Kenichi,

Abi and I talked about your suggestion in the office and below is the conclusion we arrived at:

- The way you suggested to manage the geometry is preferred; 
- The SQL script in this PR is the result of mysqldump. It's more like a backup of the schema instead of a way to manage the geometry. I will remove this script from the PR;
- `PHG4E1039InsensDetector` is currently only responsible for the FMag (and part of the E906 shielding), KMAG and absorber. They will not move the the foreseeable future thus it makes better sense to create a ROOT macro to handle the geometry construction, like we did for target and collimator;
- In the meantime, since this PR addresses some important bugs in the current geometry setup. I would suggest we merge this PR as a temporary workaround while we work on the aforementioned ROOT macros. This class should be removed once the ROOT macros are ready.

PR #62, opened 2019-10-30
Additions of utility functions and parameters.

knakano0524:
I added `SQParamDeco`, which is a data node to record a set of decoding parameters (such as channel mapping) onto DST.  This info is available in recently-decoded DSTs and also MySQL DB since `DbUpRun` uploads it at the end of decoding.  Here is a viewer;
http://e906-gat6.fnal.gov:8081/data-summary/e1039/deco-status.php

I created a new directory, `packages/UtilAna`, and put existing and new utility classes in it.  They are used by the decoder and the OnlMon clients for now, but I expect they will be used by the tracker and user-level analyses.

I included `/data2/e1039/exec-decoder.sh` in this repository after some modifications. It enables anyone to execute the decoder in a development mode, which does not overwrite the outputs made by the official decoding process.  I expect subsystem experts use this to test the OnlMon code.


online/decoder_maindaq/DecoStatusDb.cc
```diff
@@ -1,15 +1,15 @@
 #include <iostream>
 #include <TSQLServer.h>
 #include <db_svc/DbSvc.h>
+#include <UtilAna/UtilOnline.h>
 #include "DecoStatusDb.h"
 using namespace std;
 
 DecoStatusDb::DecoStatusDb() :
-  m_name_schema("user_e1039_maindaq"),
```
knakano0524: It was a bug.  The schema name must be managed by UtilOnline.

packages/PHGeometry/CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ add_library(phgeom_io SHARED
 
 target_link_libraries(phgeom_io -lphool -L${ROOT_PREFIX}/lib -lGeom)
 
-target_link_libraries(phgeom -lphgeom_io -lSubsysReco)
+target_link_libraries(phgeom phgeom_io -lSubsysReco)
```
knakano0524: This update makes the dependency correct.  In the past `phgeom` was sometimes compiled earlier than `phgeom_io` and failed at linking to it.

knakano0524:
The updated code is already running fine on seaquestdaq01 to decode new runs.  So I think it sufficient that Abi would just go through the concept (meaning) of each change.  But, if he likes to execute the updated code for test, I can provide a test procedure.

knakano0524:
I'm sorry for many commits after the pull request.  But they (particularly the last one) will definitely resolve the problem that the OnlMon server (and thus the decoder) dies when someone uses the OnlMon viewer before the 1st BOS event happens.

PR #63, opened 2019-12-06
E906legacygen abi

abinashpun:
In this pull request,
## Legacy Generators
I have created E906LegacyGen in e1039-core/generators and some details are in the READ.md file:

https://github.com/abinashpun/e1039-core/blob/E906legacygen_abi/generators/E906LegacyGen/README.md

## To read the information in module_example
You need to add a node to your module (for eg. in module_example) to read the truth dimuon information along with the cross-section for each event. I can give a separate pull request for the module_example area if necessary. But we do need to add the following line in the process event of TrkEvl.cxx and add  SQPrimaryGen library in traget_link_libraries{} of the CMakeLists.txt

image

## Vertex Generator added in PHGParticleGun and PHG4SimpleEventGenerator
There have been changes to use the SQPrimaryVertexGen for the above-mentioned generator. 

##Fun4All macro change
We have to add the library associated with E906LegacyGen (libSQPrimaryGen.so) in the macro and register the legacy classes in the macro to use it. Also, we can use switches to the current particle gun and simple event generator to use the legacy vertex generator.

I plan to give a separate pull request for theFun4Sim.C to use the features of E906Legacygen


knakano0524:
I tried to build this version on `spinquestgpvm01` and encountered a cross-dependency problem.  I think some files in `generators/E906LegacyGen` (at least `SQPrimaryVertexGen.*`) have to be moved to `simulation/g4main`.

When I ran `./build.sh` and got the following error in `simulation/g4main`.
```
input_line_12:271:11: fatal error: 'E906LegacyGen/SQPrimaryVertexGen.h' file not found
#include  <E906LegacyGen/SQPrimaryVertexGen.h>
```

Thus I ran `./build.sh generators/E906LegacyGen` and got the following error.
```
input_line_12:24:10: fatal error: 'g4main/PHG4ParticleGeneratorBase.h' file not found
#include <g4main/PHG4ParticleGeneratorBase.h>
```

Each directory needs a file in the other.  This cross dependency has to be resolved somehow.  Also `build.sh` should include the new directory `generators/E906LegacyGen` at a proper line (i.e. order).

knakano0524:
Thank you for having merged my pull request #1.  I think you can now merge this pull request.  I myself have not confirmed that the new version runs fine, but I think to do that after the merge.  We better first establish the build procedure (i.e. PR 64) so that you, I and all others build and test the new version.

PR #64, opened 2019-12-11
Online bug fix.  Better handling of environmental variables.

knakano0524:
I first fixed a problem (bug) that happens when a run fails at starting. It happened in runs 1065, 1067 and 1069.  This updated version is already running on seaquestdaq01, and the problem never happens since then.

I updated `script/this-core-org.sh`.  It now cleans up the environmental variables (like `PATH` and `ROOT_INCLUDE_PATH`) when `E1039_CORE` is already defined.  The similar change has been made on `this-share.sh` in the e1039-share package.  As a result, we can change the E1039_* environment at any time, by executing `this-core.sh` or `this-e1039.sh`.  This update will help Abi and me to test PR 63 (and future updates).

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,29 +32,27 @@ git clone git@github.com:E1039-Collaboration/e1039-core.git
 
 ## Build and Install
 
-The following commands should succeed on the supported servers.
-You should first try them to learn the overall procedure.
+You can first try the following commands to learn the overall procedure.
+They should succeed on the supported servers.
 ```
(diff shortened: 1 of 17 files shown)


abinashpun:
Need to include the E906LegacyGen in build.sh

knakano0524:
> Need to include the E906LegacyGen in build.sh

It has been included in the master branch via PR 63.  So it is not necessary here.

PR #65, opened 2019-12-18
Added some interface for the E906LegacyGen

abinashpun:
Added interfaces:
-beam profile for SQPrimaryVertexGen  <br />
-xF and m range for the SQPrimaryParticleGen  <br />

Also, fixed the bug related to the LHAPDF in CMake file

### DIFF
--- a/generators/E906LegacyGen/SQPrimaryVertexGen.C
+++ b/generators/E906LegacyGen/SQPrimaryVertexGen.C
@@ -15,22 +15,18 @@ from Kun to E1039 experiment in Fun4All framework
 #include <TFile.h>
 
 #include "SQPrimaryVertexGen.h"
-//#include "SQBeamlineObject.h"
-double beam_vtx_x; 
-double beam_vtx_y;
+
  TF2* beam_global;
+
 SQPrimaryVertexGen::SQPrimaryVertexGen():
  _beam_profile(nullptr)
 {
-inited = false;
- // beam_vtx_x = 0;
- // beam_vtx_y = 0;
- 
+inited = false; 
 }
 
 SQPrimaryVertexGen::~SQPrimaryVertexGen()
 {
-	delete _beam_profile;
+  if (_beam_profile) delete _beam_profile;
 	
   return;
 }
@@ -39,10 +35,7 @@ SQPrimaryVertexGen::~SQPrimaryVertexGen()
 int SQPrimaryVertexGen::InitRun(PHCompositeNode* topNode){ 
 
   if(_beam_profile){
-     double x, y;
-     beam_global = get_beam_profile();
-     beam_global->GetRandom2(x, y);
-  
+     beam_global = get_beam_profile();  
    }
   return 0;
 
--- a/generators/E906LegacyGen/SQPrimaryVertexGen.C
+++ b/generators/E906LegacyGen/SQPrimaryVertexGen.C
@@ -16,33 +16,38 @@ from Kun to E1039 experiment in Fun4All framework
 
 #include "SQPrimaryVertexGen.h"
 //#include "SQBeamlineObject.h"
-
-SQPrimaryVertexGen::SQPrimaryVertexGen() 
+double beam_vtx_x; 
+double beam_vtx_y;
+ TF2* beam_global;
+SQPrimaryVertexGen::SQPrimaryVertexGen():
+ _beam_profile(nullptr)
 {
 inited = false;
-
-}
-SQPrimaryVertexGen::~SQPrimaryVertexGen(){}
-
-void SQPrimaryVertexGen::Initfile(){
+ // beam_vtx_x = 0;
+ // beam_vtx_y = 0;
  
 }
 
-void SQPrimaryVertexGen::InitRun(PHCompositeNode* topNode)
+SQPrimaryVertexGen::~SQPrimaryVertexGen()
 {
-  beam_profile= false;
-  beamProfile = NULL;
-  beamProfile = new TF2("beamProfile", "exp(-0.5*(x-[0])*(x-[0])/[1]/[1])*exp(-0.5*(y-[2])*(y-[2])/[3]/[3])", -10., 10., -10., 10.);
-  beamProfile->SetParameter(0, 0.0);
-  beamProfile->SetParameter(1, 0.68);
-  beamProfile->SetParameter(2, 0.0);
-  beamProfile->SetParameter(3, 0.76);
-
-  // beamProfile = new TF2("beamProfile", "exp(-0.5*(x-0.)*(x-0.)/0.414/0.414)*exp(-0.5*(y-0.)*(y-0.)/0.343/0.343)", -10., 10., -10., 10.);
-  
-  nPieces=0;
- 
+	delete _beam_profile;
+	
+  return;
 }
+
+
+int SQPrimaryVertexGen::InitRun(PHCompositeNode* topNode){ 
+
+  if(_beam_profile){
+     double x, y;
+     beam_global = get_beam_profile();
+     beam_global->GetRandom2(x, y);
+  
+   }
+  return 0;
+
+}
+
 
 void SQPrimaryVertexGen::traverse(TGeoNode* node,  double&xvertex,double&yvertex,double&zvertex) 
 {
@@ -55,7 +60,6 @@ void SQPrimaryVertexGen::traverse(TGeoNode* node,  double&xvertex,double&yvertex
   generateVtxPerp(x, y);
 
 
- 
   for(int i = 0; i < node->GetNdaughters(); ++i) // Loop over daughter volumes
     {
      
@@ -281,15 +285,18 @@ void SQPrimaryVertexGen::traverse(TGeoNode* node,  double&xvertex,double&yvertex
 
 void SQPrimaryVertexGen::generateVtxPerp(double& x, double& y)
 {
-  if(beam_profile)
+  
+  beamProfile = ::beam_global;
+  if(beamProfile)
+  { 
+  beamProfile->GetRandom2(x, y);
+  }
+  else
   {
-       beamProfile->GetRandom2(x, y);
-    }
-    else
-    {
-      x=gRandom->Gaus(0.,0.414);
-      y=gRandom->Gaus(0.,0.343);
-    }
+    x=gRandom->Gaus(0.,0.414);
+    y=gRandom->Gaus(0.,0.343);
+  }
+
 }
 
 void SQPrimaryVertexGen::findInteractingPiece()
@@ -298,4 +305,5 @@ void SQPrimaryVertexGen::findInteractingPiece()
   index = TMath::BinarySearch(nPieces+1, accumulatedProbs,gRandom->Uniform(0,1));
  
 }
+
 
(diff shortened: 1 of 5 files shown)


generators/E906LegacyGen/SQPrimaryVertexGen.C
```diff
@@ -16,34 +16,39 @@ from Kun to E1039 experiment in Fun4All framework
...
-void SQPrimaryVertexGen::InitRun(PHCompositeNode* topNode)
+SQPrimaryVertexGen::~SQPrimaryVertexGen()
 {
-  beam_profile= false;
-  beamProfile = NULL;
-  beamProfile = new TF2("beamProfile", "exp(-0.5*(x-[0])*(x-[0])/[1]/[1])*exp(-0.5*(y-[2])*(y-[2])/[3]/[3])", -10., 10., -10., 10.);
-  beamProfile->SetParameter(0, 0.0);
-  beamProfile->SetParameter(1, 0.68);
-  beamProfile->SetParameter(2, 0.0);
-  beamProfile->SetParameter(3, 0.76);
```
knakano0524: As I see these lines were removed, you seem to stop using "beamProfile".If so, the variable "beamProfile" itself should be removed from this class.

generators/E906LegacyGen/SQPrimaryVertexGen.C
```diff
@@ -16,34 +16,39 @@ from Kun to E1039 experiment in Fun4All framework
...
-
-  // beamProfile = new TF2("beamProfile", "exp(-0.5*(x-0.)*(x-0.)/0.414/0.414)*exp(-0.5*(y-0.)*(y-0.)/0.343/0.343)", -10., 10., -10., 10.);
+	delete _beam_profile;
```
knakano0524: if (_beam_profile) delete _beam_profile;

generators/E906LegacyGen/SQPrimaryVertexGen.h
```diff
@@ -16,23 +16,21 @@ from Kun to E1039 experiment in Fun4All framework
...
 #include <TVector3.h>
 #include <TH1F.h>
+#include <fun4all/SubsysReco.h>
 #include "SQBeamlineObject.h"
 
 class PHCompositeNode;
 class SQBeamlineObject;
 
-class SQPrimaryVertexGen
+class SQPrimaryVertexGen: public SubsysReco
```
knakano0524: This class became a SubsysReco child class here, but it is not complete since its process_event() function hasn't been defined.  How do you intend to use this class?

abinashpun: I intend to make this SubsysReco child class because I wanted to set the beam profile in the Fun4Sim macro. So it needs to be registered into the Fun4All server.

abinashpun: The Fun4Sim line is 
https://github.com/E1039-Collaboration/e1039-analysis/blob/c697b092036563e1dd0afb6cb09385da70ff26f8/SimChainDev/Fun4Sim.C#L76-L79

generators/E906LegacyGen/SQPrimaryVertexGen.C
```diff
@@ -16,34 +16,39 @@ from Kun to E1039 experiment in Fun4All framework
 
 #include "SQPrimaryVertexGen.h"
 //#include "SQBeamlineObject.h"
-
-SQPrimaryVertexGen::SQPrimaryVertexGen() 
+double beam_vtx_x; 
+double beam_vtx_y;
+ TF2* beam_global;
```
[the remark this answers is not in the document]
abinashpun: I will delete the beam_vtx_x and beam_vtx_y (Thank you)

I defined TF2* beam_global so that I can pass the profile set from Fun4Sim to https://github.com/abinashpun/e1039-core/blob/c357feb85eb1fbb188cd5c3b078b7b2852a8ee28/generators/E906LegacyGen/SQPrimaryVertexGen.C#L39-L46
and can be passed to another function generateVtxPerp() as 
https://github.com/abinashpun/e1039-core/blob/c357feb85eb1fbb188cd5c3b078b7b2852a8ee28/generators/E906LegacyGen/SQPrimaryVertexGen.C#L286-L293

generators/E906LegacyGen/SQPrimaryVertexGen.C
```diff
@@ -16,34 +16,39 @@ from Kun to E1039 experiment in Fun4All framework
...
+  return;
+}
+
+
+int SQPrimaryVertexGen::InitRun(PHCompositeNode* topNode){ 
+
+  if(_beam_profile){
+     double x, y;
+     beam_global = get_beam_profile();
+     beam_global->GetRandom2(x, y);
```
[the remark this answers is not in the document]
abinashpun: Yes, I will remove the line 44 (Thank you)

generators/E906LegacyGen/SQPrimaryVertexGen.C
abinashpun: I will implement that..thank you

generators/E906LegacyGen/SQPrimaryVertexGen.C
abinashpun: I have used it in the generateVtxPerp() function.
https://github.com/abinashpun/e1039-core/blob/c357feb85eb1fbb188cd5c3b078b7b2852a8ee28/generators/E906LegacyGen/SQPrimaryVertexGen.C#L289-L292

generators/E906LegacyGen/SQPrimaryVertexGen.C
```diff
@@ -16,34 +16,39 @@ from Kun to E1039 experiment in Fun4All framework
 
 #include "SQPrimaryVertexGen.h"
 //#include "SQBeamlineObject.h"
-
-SQPrimaryVertexGen::SQPrimaryVertexGen() 
+double beam_vtx_x;
```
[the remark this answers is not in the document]
abinashpun: Thank you, I will remove them

generators/E906LegacyGen/SQPrimaryVertexGen.h
knakano0524: It can be registered but does nothing per event because its process_event() is not implemented.  Saying in other words, what calls "SQPrimaryVertexGen::traverse()" when?

generators/E906LegacyGen/SQPrimaryVertexGen.C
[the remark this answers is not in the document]
knakano0524: The two blocks that you mentioned are in the member functions.  Thus a member variable (beamProfile or _beam_profile) should be sufficient.

generators/E906LegacyGen/SQPrimaryVertexGen.C
knakano0524: I thought "beamProfile" was always "null" (without "= new TF2") and thus "if(beamProfile)" is always "false".  But now I see that you moved this part to "Fun4Sim.C".

What is an advantage of new'ing "beam_profile" in Fun4Sim.C?  I think we better keep such technical object inside the class.

generators/E906LegacyGen/SQPrimaryVertexGen.h
abinashpun: Yes, it is not doing anything per event. I just use it to set the beam-profile at the beginning of the run via IntRun() function of Subsysreco.

The SQPrimaryVertexGen::traverse() is called in the process_event of SQPrimaryParticleGen for setting up the vertices.

generators/E906LegacyGen/SQPrimaryVertexGen.C
abinashpun: I came up with this idea because I think we don't know the exact beam profile. So, in the future, we can set it simply via Fun4Sim without digging up the e1039-core area. Please let me know what you think about it.

generators/E906LegacyGen/SQPrimaryVertexGen.C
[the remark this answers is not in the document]
abinashpun: I tried different ways not to use any global variable. I tried it in that way too but could not make it work. Maybe I can check again with other ideas.

generators/E906LegacyGen/SQPrimaryVertexGen.C
knakano0524: Yes, the parameters should be adjustable.  I think a simpler member function can do that such as "SQPrimaryVertexGen::setProfile(x_center, y_center, x_width, y_width)".

I expect we won't change the functional shape itself (i.e. Gaussian).  Thus user need not handle TF2 directly.

generators/E906LegacyGen/SQPrimaryVertexGen.h
knakano0524: Since it is not the standard usage of SubsysReco module, the current code is (and will be) confusing.  We can rather make SQPrimaryVertexGen a member of SQPrimaryParticleGen, and initialize it in SQPrimaryParticleGen::InitRun().

generators/E906LegacyGen/SQPrimaryVertexGen.C
abinashpun: Ok. But I think we also expect to have some tail function in addition to Gaussian. I can try as you suggested. But for now, I came up with this convention. 
https://github.com/E1039-Collaboration/e1039-analysis/blob/558faecc40a62186f62db8310af1206ef46e7949/SimChainDev/Fun4Sim.C#L75-L91

Let me know what you think about it.

generators/E906LegacyGen/SQPrimaryVertexGen.h
abinashpun: The only reason for making it Subysreco class is to initialize the beam profile for now. I am thinking of this vertex generator code to make stand alone so that it can be used in the other minimum bias generator for eg. particle guns, simple event generators in the future.

generators/E906LegacyGen/SQPrimaryVertexGen.C
knakano0524: So I would agree to keep the current interface, as you expect soon(?) to change the function shape.  I thought such change was rare.

generators/E906LegacyGen/SQPrimaryVertexGen.h
knakano0524: I understand the purpose (making it standalone) and agree it is better.  But I mean the method is not suitable at present in several lines of this class.

If this class is SubsysReco, it should have its own "process_event()" (which generate a vertex position in data node).  It should not be called by another SubsysReco like SQPrimaryParticleGen.

knakano0524:
I looked into SQPrimaryVertexGen::traverse().  It seems to me that lines 56-247 need not be repeated in every event;
https://github.com/abinashpun/e1039-core/blob/legacy_add/generators/E906LegacyGen/SQPrimaryVertexGen.C#L56-L247
This line block counts up the amount of materials, which is constant over events.  If so, it should be done once in the initialization.

abinashpun:
> I looked into SQPrimaryVertexGen::traverse(). It seems to me that lines 56-247 need not be repeated in every event;
> https://github.com/abinashpun/e1039-core/blob/legacy_add/generators/E906LegacyGen/SQPrimaryVertexGen.C#L56-L247
> This line block counts up the amount of materials, which is constant over events. If so, it should be done once in the initialization.

I think that is true. I can try to come up with more efficient way later.

generators/E906LegacyGen/SQPrimaryVertexGen.h
abinashpun: Yes, I think I agree with you for most of the part. It would be better if I can make the vertexgen code more efficient and move it to the another location in such a way that the particle guns and other event generators can also use it. I will keep on working on it but for Monday, I just want to keep as it is (if you are ok with it).

PR #73, opened 2020-04-12
3 small changes to the build script which I found useful in the development

liuk:
1. added a -i option to allow build incrementally. It is essentially the -s option without deleting the existing build directory;
2. added a -c option to pass additional options to the cmake
3. for all build modes, the script will attempt to delete the installation files from the previous build to guarantee the new headers/libraries will be installed

changes 1 and 2 do not change the existing behavior; change 3 does change the existing behavior of the script but should be transparent to user.


knakano0524:
All look good.  A comment on change 3 is that "install_manifest.txt" does not include pcm files and thus the deletion is not complete for now.  To make it complete, we should modify "CMakeLists.txt" somehow.

PR #76, opened 2020-05-09
Integration of GenFit into the kTracker reconstruction

liuk:
This is a fairly big change on the reconstruction part of the code. The primary purpose is to provide an alternative track fitting algorithm under the hood to replace the original Kalman filter we inherited from E906. The work was initiated by Haiwang but so far only the track extrapolation has been using the GenFit code, the rest of covariance calculation, KF algorithm, etc. are still based on the E906 program. This change implemented a wrapper for the GenFit to merge with our current reconstruction package. 

The detailed major changes are:
1. re-named packages/ktracker to packages/reco  and re-structured the sub-directories under ktracker. Now we have:
      - **reco/interface**: this directory contains all the classes as data objects which can be saved in a TTree or Fun4All node tree;
      - **reco/kfitter**: this directory contains all the classes related to the old E906-style Kalman filter-based fitter;
      - **reco/SQGenFit**: this directory contains the wrapper and interface to bridge E1039 reconstruction with GenFit
      - **reco/ktracker**: this directory contains all the Fun4all modules to be called by Fun4All server, as well as their supporting classes;
      - _reco/patterndb_: this directory contains Haiwang's work of dictionary-search based track finding. Since we do not have man power to develop or maintain this code, it's temporarily moved to its own directory and excluded from the build script;

2. A new Fun4All module called SQReco is introduced as a wrapper of the track finding and fitting process. User can select different data types (E906 or E1039) and different fitter choices (old E906 fitter, or GenFit fitters) via a series of setter functions. An example macro of running over E906 Run6 data is provided under packages/reco/ktracker/test/test_SQReco.C. We expect this module to replace the current KalmanFastTrackingWrapper in all the example macros;

3. Even though we plan to deprecate the KalmanFastTrackingWrapper, it is still maintained at this time for backward compatibility. There are two major things to mention regarding KalmanFastTrackingWrapper:
      - under the hood KalmanDSTrk is replaced with KalmanFastTracking. These two are identical if the user turns off the dictionary search, i.e. set DS level to 0;
      - the setters/getters related to KalmanDSTrk are still kept in KalmanFastTrackingWrapper so that it won't break user's macro (for example the current version of Fun4Sim.C), although they will do nothing

4. build.sh is updated accordingly.

To make sure the changes are backward compatible, I run the out-of-the-box Fun4Sim.C from E1039-analysis for 1000 events and the reconstruction performance stays the same, as they should be since nothing really changed. Below are two figures of the number of tracks found per event and number of hits found per track. In both figures the blue solid hist is from PR75 and red dashed hist is from this PR. The performance of the new GenFit-based fitter will require further study with more user's input.

Snipaste_2020-05-09_12-35-31Snipaste_2020-05-09_12-40-34


liuk:
I added a new analysis directory under `e1039-analysis` called `RecoDev` to show two example macros to run the new `SQReco` with both E906 data and E1039 simulation.

knakano0524:
Thank you for working on this big change, which should be helpful for future developments.  I have run my eyes through the changes and tested a full build on gpvm01.  All look fine.  I think I myself need not run the reconstruction since Kun already has done that.

PR #79, opened 2020-05-19
Update the geometry construction for spectrometers

liuk:
eThis PR aims to improve the style of geometry construction for the spectrometer. The primary purpose is to standardize the geometry construction macros and get rid of old E1039InsenseDetectors which has a special geometry schema hard-coded in.

### Major changes in the PR includes:
- A modification of DbSvc to add support to SQLite-based geometries, which is essentially can be regarded as a file-based configuration file. The current geometry definition of FMag/KMag are stored in `$E1039_RESOURCE/geometry/magnetic_fields/magnet_geom.db`. 
- A new set of g4subsystem class to facilitate the geometry construction from a geometry sqlite db file. The original E1039InsensSubsystem is still untouched and kept but should be removed in the future.
    1. `SQG4DipoleMagnetSubsystem`: standard interface to PHG4Reco;
    2. `SQG4DipoleMagnetSteppingAction`: not really needed for inactive volumes. But for FMag/beamdump, it is useful (and implemented for E906) to use stepping action to prune the low-energy tracks and tracks with large slopes (will go out of acceptance). This feature is turned off by default and can be enabled by `set_int_param("enable_track_filter", 1)` for FMag;
    3. `SQG4DipoleMagnetDetector`: this class implements the geometry by reading from the sqlite db file passed on by `set_string_param("geomdb", $E1039_RESOURCE/geometry/magnetic_fields/magnet_geom.db)`;
- Added a lot of material definition to the end of `PHG4Reco::DefineMaterials` since this apparently is how PHENIX handles materials;
- Updated the geometry setup macros in `$E1039_CORE/macros`
    1. added a new macro `G4_InseneitiveVolumes.C` to construct Shielding, FMag, KMag and MuID absorber;
    2. added a new macro `G4_Beamline.C` to construct beamline objects, so far only collimator is implemented;
    3. removed the collimator and shielding from `G4_Target.C` to the two new macros above;
    4. added a macro `Fun4DumpGeom.C` to construct and dump the geometry for debugging;
- Included `$E1039_CORE/macros` in the `$ROOT_INCLUDE_PATH` so that all analysis macros can include the standard geometry macros using lines like `#include <G4_Target.C>`, instead of keeping a local copy of the macro in the same directory.

### Before/after the proposed change
Since this PR merely added new classes to address the geometry of FMag, KMag and MuID absorber, we can compare the exported geometry using `E1030InsensSubsystem` and `G4_InsensitiveVolumes.C`. Below is the screenshot, the detailed export shows they are identical.

Geometry using `E1039InsensSubsystem`:
67327210-31aecd80-f4dd-11e9-856f-c30f5300f283

Geometry using `G4_InsensitiveVolumes.C`
Snipaste_2020-05-19_09-55-52

### Non-critical problems discovered and/or to-do lists:
During the development of this PR, I noticed several non-critical problems in our existing code. I didn't try to address them but we should keep them in mind and implement fixes in the near future:
- It seems some of our current material definitions share the same name as those defined in `PHG4Reco::DefineMaterials()` thus triggered a warning since Geant4 keeps a global map of materials created, it is not a problem, but we should follow the same material convention;
- There are some non-critical overlaps on the edges of both FMag and KMag, the same overlap existed in E1039InsensSubsystem and also in E906 GMC code. I assume it is not a problem, but we should investigate future like doing a material scan;
- We do not currently have concrete shielding downstream of FMag and before station-1. Again it is the same case in previous implementation and E906, we should get updated shielding configurations and painstakingly get them implemented;
- It seems there is a typo in `G4PhysicalVolume (G4PVPlacement)` construction in a lot of existing geometry code. The correct interface should be `G4PVPlacement::G4PVPlacement(.., G4bool pMany, G4int pCopyNo, G4bool pSurfChk=false)`, but at a lot of places `pMany` and `pCopyNo` are swapped. This is not a problem for now since C++ converted 0 and false to false and 0, which is fine but will trigger a problem if we have more than one layer of the volumes;


knakano0524:
Thank you for this very helpful arrangement.  I have run my eyes through the changes and tested a full build on gpvm01. I would ask Abi (or Kun) to merge this PR once agreeing to do so.

I wonder how we will use SQLite and local MySQL server together with the existing on-network MySQL servers in the future.  It is not urgent but we better clarify the strategy to use multiple DB types.  Otherwise we might have difficulties in organizing the location and the consistency of DB variables.

liuk:
Hi Kenichi, thanks a lot for taking the time going over the proposed changes. It seems Abi already merged this PR, but I would like to follow up on the discussion about the DB usage.

The change adding local MySQL server is entirely for the developers so that one can completely work on a local environment.  I was thinking about implementing a mariadb in the docker container with all the geometry information but eventually gave up because like you said it's probably too confusing. I'm OK with removing the localhost support since it takes a lot of extra work to make sure selecting localhost would have the same behavior.

Regarding sqlite, my intention is to use it like a local configuration file (similar to what we are using for alignment parameters). It just happens that ROOT abstracted all the SQL interface rather nicely so that the DbSvc could be directly used to read from sqlite with some very minor changes. I did plan to add an entry in the run_range table to point to this sqlite db file so that it's not explicitly specified in the macro, but didn't figure out what is the best way to implement it yet. Since the sqlite file only contains FMag, KMag and absorber which should not move during entire lifetime of the experiment, I guess it's not really critical and can probably be just hard-coded.

Please let me know if this agrees with your vision about how to manage the geometry versioning. I think Zulkaida is also working on implementing the target geometry so it would be nice if we finalize on the plan and let him know what's needed soon.

> Thank you for this very helpful arrangement. I have run my eyes through the changes and tested a full build on gpvm01. I would ask Abi (or Kun) to merge this PR once agreeing to do so.
> 
> I wonder how we will use SQLite and local MySQL server together with the existing on-network MySQL servers in the future. It is not urgent but we better clarify the strategy to use multiple DB types. Otherwise we might have difficulties in organizing the location and the consistency of DB variables.

knakano0524:
Hi Kun, thank you so much for the explanations.

Let's keep the localhost support.  It should be surely helpful for development, and only few people take care of it for now.

A simple split of the DB usage would be
 * Using MySQL for rather _dynamic_ variables, which are controlled via the run_range table, and
 * Using SQLite for rather _static_ variables, which are hard-coded in G4_*.C macro.

Do you think we can categorize all variables into these two?  Or, it should not be very difficult to implement the run_range table for SQLite, but it seems not better than just always using MySQL. 

If we try to make our software completely standalone (i..e. running on single computer without network), we should stop using MySQL.

liuk:
Hello Kenichi, thanks for the clarification this is exactly how I imagined the separation should work. For the dynamic variables it takes a lot of work and care to make them correct, and the impact of the incorrect settings might be very subtle. Therefore I would also prefer to keeping the numbers and versioning centralized, i.e. on a mysql server that only experts can update.

PR #81, opened 2020-06-02
added a template module based on the module_example to make creation of a new class easier …

IevgenLavrukhin:
Probably needs to be cleaned better but I am not confident yet in removing things from *cxx.
It was pain for me to create a new class.
Please check if you like it. And I should probably clean this template more.

### DIFF
--- a/template/CMakeLists.txt
+++ b/template/CMakeLists.txt
@@ -1,0 +1,51 @@
+# Setup the project
+cmake_minimum_required(VERSION 2.6 FATAL_ERROR)
+project(module_example CXX C)
+
+# ROOT dict generation
+file(GLOB dicts "")
+file(GLOB LinkDefhs ${PROJECT_SOURCE_DIR}/*LinkDef.h)
+foreach(LinkDefh ${LinkDefhs})
+	string(REPLACE "LinkDef.h" ".h" Dicth ${LinkDefh})
+	string(REPLACE "LinkDef.h" "_Dict.C" DictC ${LinkDefh})
+	string(REPLACE "${PROJECT_SOURCE_DIR}/" "" DictC ${DictC})
+	list(APPEND dicts ${DictC})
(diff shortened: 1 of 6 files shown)


knakano0524:
Thank you for exploiting the E1039 software framework.  

Our present software policy (or at least I) supposes that user-level analysis is carried out via `e1039-analysis`, namely [e1039-analysis/AnaRealDst](https://github.com/E1039-Collaboration/e1039-analysis/tree/master/AnaRealDst) for real data and [e1039-analysis/AnaSimDst](https://github.com/E1039-Collaboration/e1039-analysis/tree/master/AnaSimDst) for simulated data.  Do they fit for your analysis environment?  We can import `init.sh` into them, where I myself am renaming the class via editor replace function.

`e1039-core` is supposed to be system-level software.  Usually analyzer need not build it.  If you see your analysis module has to be built inside `e1039-core`, we should reconsider the policy.

IevgenLavrukhin:
HI,

1) I created this template because it is painful to create a new package every time people needs to do it. I thought it would be more convenient to do it this way.

2) Also, From what I  understand "e1039-analysis" repo contains only root macros. All the libraries, packages and new classes should be defined in e1039-core. Or I am missing anything?

Thanks,

knakano0524:
`e1039-analysis/AnaRealDst` and `e1039-analysis/AnaSimDst` contain not only macros but also source files and shell scripts in order to build user-level libraries and execute them.

I expect that each user makes a copy of AnaReal(Sim)Dst per analysis topic, as a directory or a Git branch, so that he/she can use the existing SubsysReco module, `src/AnaRealDst.cc`, as template.

liuk:
Hello Ievgen, the modules/packages in the e1039-core typically involves multiple files/classes and custom cmakelist files based on the specific needs of the package. It seems pretty difficult to generalize the behavior of starting a new package. 

From what I can see, the template and shell script is more suitable for creating analysis modules for users, which should be created inside e1039-analysis repo.

IevgenLavrukhin:
Hi, Kun,

I thought that by default Fun4All package class has 5 function:
Init, InitRun, process_event, EndRun, End. 
This is what I found in the "module_example" and created a template pattern for this so people don't need to create this on their own. 

Anyway, please discard this merge request if you think it is useless for future. 

Thanks,

PR #83, opened 2020-06-09
Removal of module_example

knakano0524:
This update removes `module_example/` because it is obsolete and confusing to new comers.

The directory contains the following SubsysReco modules; `ReadMySql`, `SimpleReadMySql`, `TestAnalyzer`, `TestSimAnalyzer` and `TrkEval`.  As they are used in some macros under `macros/`, those macros are removed also.  The e1039-analysis package will be updated in [PR 9](https://github.com/E1039-Collaboration/e1039-analysis/pull/9) so as not to use these modules.

It is possible that someone is using `TrkEval` since it was enabled by default in many e1039-analysis macros.  Its output is `trk_eval.root`.  It has two functions to create two output trees.  Here I summarize their main operations.

1. TrkEval::RecoEval(), lines 104-401
    * lines 150-187: Count N of SQ hits per true track.
    * lines 208-253: Count [N of SQ chamber hits per true track] per reco track.
    * lines 255-284: Find the best-overlapped true track for each reco track (-> "recID_bestParID").  Note: "parID_bestRecID" is never used later.
    * lines 287-391: Save the info of each reco track together with the best-overlapped true track.

2. TrkEval::TruthEval(), lines 404-
    * lines 449-506: Count N of SQ hits per true track.
    * lines 509-544: Save the true position of every SQ hits.
    * lines 558-604: Count [N of SQ chamber hits per true track] per reco track.
    * lines 606-627: Find the best-overlapped reco track for each true track (-> "parID_bestRecID").   Note: "recID_bestParID" is never used later.
    * lines 633-656: Save the kinematics of each true track at vertex.
    * lines 659-686: Extract and save the kinematics of each true track at St. 1.
    * lines 712-749: Save the info of the best-overlapped reco track per true track.
    * lines 753-834: Make and save pairs of true tracks

Most operations are duplicated in the two functions, which is why it is hard for us to keep maintaining this module.  I see the key information that this module extracts is
1. The positions and momenta of true tracks at St. 1 (and 3), and
1. The correspondence between true track and reco. track.

The 1st info is now extracted by `simulation/g4dst/TruthNodeMaker.cc` and stored in DST.  Analyzer can access it via `SQTrack::get_pos_st1()` etc. 

The 2nd info will not be available in DST.  I suppose analyzer can find it by comparing the momentum between true track and reco. track.  Note that TrkEval compares the hit pattern between true track (w/G4Hit) and reco. track (w/SQHit), but I think SpinQuest does not need such detailed comparison because of low track multiplicity.


liuk:
Hello Kenichi, thanks a lot for making this change. I compiled this code on my docker image and confirmed it works with PR9 of `e1039-analysis` out-of-the-box without a problem. The only thing I would like to mention is line 194 of `ThruthNodeMaker.cc` where the truth momentum needs to be converted from MeV to GeV which is inconsistent with the rest of the unit definitions. This is a bug uniquely in the `PHG4BlockSubsystem`, which I will fix with a PR later. For now I suggest we keep it this way and I will change both files in my PR.

PR9 of `e1039-analysis`is also tested and shows no problem. So we should go ahead and merge both PRs. Since the order might be delicate so I will leave it to you to actually click the merge button.

abinashpun:
Dear Kenichi, I also tested both of the PRs and worked fine. As I mentioned in the last meeting,  the TruthNodeMaker doesn't work with the legacy generator (neither for multiparticle gun). This is because TruthNodeMaker requires an event to be HepMC event. I am still working to make the legacy generator work within this framework. Hopefully, I can make a PR within this week.

liuk:
> Dear Kenichi, I also tested both of the PRs and worked fine. As I mentioned in the last meeting, the TruthNodeMaker doesn't work with the legacy generator (neither for multiparticle gun). This is because TruthNodeMaker requires an event to be HepMC event. I am still working to make the legacy generator work within this framework. Hopefully, I can make a PR within this week.

If the conflict is only with the legacy generator which is not turned on by default in the example macros, I would suggest we move ahead with the merge and fix the generator later.

knakano0524:
I would thank Kun and Abi for the careful checks.  OK, I will ask the collaboration if everyone is ready for the update, while cleaning up bugs and dirty lines.  Hopefully in a few days, I will merge this PR and update the offline software library.

About line 194 of `ThruthNodeMaker.cc`, the line has `/1e3`, which I think makes the conversion rightly.  Have you just overlooked it, or is it not right?

About `TruthNodeMaker`, one temporary solution is to make it accept the absence of the HepMC node.  In this case, `TruthNodeMaker` does not touch the SQMCEvent node and the legacy generator can/should do that.

liuk:
> About line 194 of `ThruthNodeMaker.cc`, the line has `/1e3`, which I think makes the conversion rightly. Have you just overlooked it, or is it not right?
> 

Hi Kenichi, I didn't make myself clear. This line correctly converted the MeV from PHG4Hit to GeV by /1E3, however this should not be needed as all other geometries (like target/dump, etc.) fill PHG4Hit with GeV by default. The only subsystem that uses MeV is the block subsystem which is used to describe all the detectors. I think to make everything consistent the conversion from MeV to GeV should be done in the stepping action of the block subsystem, which I plan to do in my upcoming PR.

PR #84, opened 2020-06-17
Bug fix and re-structure of the sensitive volume construction and digitization

liuk:
This is a large and potentially **not** backward-compatible change. I discovered several problematic approach in the sensitive volume construction during my attempt in implement a EMCal detector for the dark photon research.  Below is a breakdown of all the problems and changes.

## Construction of the sensitive volume for the drift chambers.

For some reason, not all drift chamber planes are implemented in the Geant4 geometry. For every primed/unprimed pairs, only the first plane is constructed, and the second plane is missing, which means for half of the chamber volume the tracks are going through air instead of the chamber gas. In the digitization stage, the `G4Hit` in the first plane is linearly extrapolated to the second plane and generate a `SQMCHit`. This is probably fine for station-3 but around station-2 where the fringe magnetic field is strong and chamber is big it might be better to let Geant4 do the track propagation instead. 

There are several other minor problems regarding the detector volume: the material is hard-coded in the geometry service and incorrect; for drift chambers the drift distance is calculated using position where the track intercepts the wire plane, instead of the distance of the closest approach between track the the hit wire; for all the detectors, the surveyed rotation are not used in the simulation.

None of the problems will generate drastically wrong result, but we should fix them anyways. To fix these problems, I significantly updated the `G4_SensitiveDetectors.C` macro and changed behavior of the digitizer. **Since the geometry construction is also used elsewhere, I'm not sure if this change will break other packages, and would like to ask Kenichi to verify before we merge the change**.

## Changes to the geometry service and digitizer

To make the aforementioned complicated digitization happen, previously we have a concept called detector groups, which is used to tell which detector's `g4hits` should be used to generate digitized hits on other detectors. This is entirely hard-coded and duplicated in two places (`GeomSvc` and `DPDigitizer`), which is no longer needed once we create a volume for each detector separately. 

Since `DPDigitizer` was ported from a small simulation project for the Dark photon physics, the class creates and initializes another set of class to hold the geometry information which are duplication of what we have in the `GeomSvc`. The initialization process is hard-coded to a specific geometry schema, which happens to be the default schema specified by the `DbSvc`. These duplication and hard-coding have all be removed and now the digitizer gets its geometry information from the geometry service. 

The changes to the digitizer are pretty significant, so I created a new digitizer class called `SQDigitizer` to replace the `DPDigitizer`, and used a `typedef` statement at the top of the G4_SensitiveVolume.C to create an alias so the macros which calls the `DPDigitizer` will be actually calling the new `SQDigitizer` instead. This is entirely a personal preference since we can just keep the old name, so I will leave it up to you guys to decide which way to go. 

## Other minor changes

- The stepping action of the PHG4BlockSubsystem did the unit conversion (mm -> cm) for the position but forgot to do it for the momentum. It is fixed in this PR. I also changed Kenichi's TruthNodeReader accordingly;
- We have both truth position and truth momentum information available in the PHG4Hit, but only the truth position info is saved in the SQMCHit_v1. I added pure virtual interfaces in SQHit and actually implementation in SQMCHit_v1 to include truth momentum as well.

## Tests

I generated 1.5k single mu+ (about 700 in the acceptance) in the simulation chain to see the overall impact of the change. Below are the comparison of the reconstruction results before and after the PR.

Snipaste_2020-06-17_11-52-22 Detector display before the fix, notice for D2 it's clear only 3 out of 6 planes are constructed.
Snipaste_2020-06-17_11-50-54 Detector display after the fix. All 6 D2 planes are visible and it's the same for D1 and D3, just less obvious from the screenshot. But one can verify the G4HIT_* nodes exist for all chamber planes now. The color change is an artifact of changing the construction material.

I also tested how reconstruction performs before/after the change, as a quick test of the entire simulation/digitization process. Below are the distribution of the chi^2, number of hits found, and residual distribution of all chambers before and after the change (Red is after the PR, blue is before). The difference of the track quality seems  very small, which is reasonable since none of the problems fixed in the PR would have a huge and obvious impact.

The chi^2 distribution.
Snipaste_2020-06-17_12-07-34 

Number of hits found on the tracks.
Snipaste_2020-06-17_12-11-01 

Residual distribution of all chambers.
Snipaste_2020-06-17_12-14-31


knakano0524:
I built this version on gpvm01 successfully.  I then ran `e1039-analysis/SimChainDev` as a simple test.  It failed due to the following error;
```
root -b -l -q 'Fun4Sim.C(100)'
...[snip]...
G4GDML: Writing setup...
PHG4GDML: Writing surfaces...
PHG4GDML: Writing '/tmp/PHGeomUtility_geom_file_479.gdml' done !
DPDigitizer: Could not locate g4 hit node G4HIT_DP1BL
/seaquest/users/kenichi/e1039/git/e1039-core/framework/fun4all/Fun4AllServer.cc:837: Module DPDigitizer issued Abort Run, exiting
```
Doesn't this error happen on your environment?  The geometry configuration seems not consistent somewhere.

After this error is resolved, I will run the event display (on simulated event) under this new geometry construction.  I will also look into the files changed.

liuk:
Hi Kenichi, thanks for trying it out. This is one of the non-backward-compatible part that I forgot to mention in the PR request. The interface of `G4_SensitiveVolumes.C` has been changed to allow use to enable/disable building DP hodo and station-1 drift chamber. The new interface now becomes:
```
SetSensitiveDetectors(PHG4Reco*, bool do_dphodo = true, bool do_station1DC = false, int verb = 0)
```
Currently in the `Fun4Sim.C` line 229, the function is called with only 2 parameters (`SetupSensitiveVolumes(g4Reco, 0)` with 0 being the verbosity level), thus DP hodo is disabled. Please change line 229 of `Fun4Sim.C` to something like `SetupSensitiveDetectors(g4Reco);`, which should work in both old and new PR using all the default value of the other parameters. I will provide a interface in the digitizer to explicitly enable/disable the 2 optional detectors in the digitization chain as well.

Once we decide to merge this PR, I will update the corresponding analysis macros.

abinashpun:
I was also getting the same error as Kenichi earlier.

I tried with the change in Fun4Sim as suggested by Kun. Now, It runs fine. Also, I can verify there are G4HIT* nodes for all the drift chamber planes.

knakano0524:
Thank you for the solution.  It works in my case as well.  I have confirmed that the event display runs fine.  After this PR is merged, I will modify `e1039-analysis/EventDisplaySim` accordingly (with additional changes).  I would ask Kun to make the merge action.

PR #85, opened 2020-07-04
Bug fix in the field map and G4 simulation performance improvement

liuk:
This PR consists of 2 parts: the first one is to fix a bug in the `PH3DFieldCartisian` class which was used for the field map interpolation in our core software; the second one is to improve the speed of G4 simulation via several changes. The original purpose of the PR was only to improve the simulation speed, however during the process I noticed the bug in the field map which caused the momentum reconstruction to be off by ~1%.

## Bug fix in the `PH3DFieldCartisian` 

In the current core software, the KMag and FMag are implemented by `packages/PHField/PHField3DCartesian`. This class read the text-based input grid file (contains x/y/z/Bx/By/Bz), save the content in a map using (x, y, z) as key and (Bx, By, Bz) as value. To obtain the magnetic field strength at a spacepoint (x, y, z), this class uses the map to locate the 8 grid points that are closest to the give spacepoint and performs a trilinear interpolation. There are two problems with this approach:
 1. using a map to locate the closest grid points, although being a o(log N) operation, is still very inefficient given that we are using fixed step size on x/y/z thus the index of the 8 closest grid points should be directly calculated using constant time;
 2. the implementation of the trilinear interpolation in `PHField3DCartesian` has a bug which have the upper-bound and lower-bound inverted. The difference is almost negligible inside the dipole magnets, but is noticeable in the fringe field region. The overall effect of this bug is that the field integral is increased by about 1%, which makes the reconstructed momentum smaller than the true momentum, as reported by Abi.

To fix this bug, I implemented another field map class `SQ3DFieldCartesian` under `package/PHField` with a hand-crafted trilinear interpolation algorithm (https://en.wikipedia.org/wiki/Trilinear_interpolation). The class `PHFieldSeaQuest` is also changed to use the new field map class.

## G4 simulation performance improvement

In addition to the more efficient field map interpolation method implemented above, this PR introduced 4 other small changes to speed up the simulation process.

  1. currently in addition to the default `"FTFP_BERT"` physics list, in the `PHG4Reco::InitRun()` optical photon process is also enabled by default, which is not useful in our simulation. I disabled this part by default and added a method `PHG4Reco::set_optical_photon(bool)` so that user can manually turn it back on;
  2. currently the simulation will track all particles till it's out-of-boundary or killed by Geant, a large amount of soft particles takes a lot of time to process. I added a function `PHG4Reco::set_energy_threshold(double)` to set the energy threshold for the G4 tracking, all the particles below this energy threshold will be killed without further propagation. This feature is turned off by default and user needs to explicitly set a threshold value to turn it on;
  3. currently we are using a global magnetic field manager even though the field map itself only covers down to z ~1550 cm. For tracks with z > 1550 cm the global field manager will always return 0 magnetic field but the RK4 propagation is still carried out for each step which generates a lot of useless computation. I created a zero-field-sub-world volume that contains all the downstream detectors with no magnetic field (all station-3 and station-4). The RK4 propagation is turned off in this volume to save time. This feature is turned off by default, user needs to call `PHG4Reco::set_zero_field_line(double)` to specify the beginning z position of the zero-field-sub-world to turn it on;
 4. currently at the beginning of the simulation, we read the text-based field map for FMag and KMag to create the look-up table for the field map interpolation. The text-based field map files are 59MB for FMag and 7MB for KMag. They take about 5 seconds to be loaded while running locally, which is not too bad. But when running tens of thousands of jobs on the grid, both field maps need to be uploaded to each node and thus constitute a significant overhead. In this PR I converted them to `TTree` and saved in a ROOT file, the size of the files are significantly reduced to 13MB and 1MB respectively. The local loading time dropped to about 1s. The old text-based field map is still supported but we should remove them from the resource package at a later stage.

## Tests

For all the tests below I generated 1500 single mu+ from `z = -300cm` with `-6GeV < px < 6GeV, -3GeV < py < 3GeV, 25GeV < pz < 100GeV`. The macro can be found here: https://github.com/liuk/e1039-analysis/blob/master/RecoDev/RecoE1039Sim.C.

### Field map bug fix

By changing to the new field map implementation, the average running time of `PHG4Reco` per event is reduced from **198ms** to **161ms**.  Below are the resolution of the momentum reconstruction at station-1 before/after the change. Left is before the change, right is after.

drawingdrawing

### Simulation speed improvement

On top of the test above, following three lines are added after line 124 of https://github.com/liuk/e1039-analysis/blob/master/RecoDev/RecoE1039Sim.C to enable the new speed improvement features in this PR:

```
g4Reco->set_optical_photon(false);   // disable the optical photon process
g4Reco->set_energy_threshold(0.0005);    // set the minimum energy threshold of particles to 0.0005GeV/0.5MeV
g4Reco->set_zero_field_line(1600);    // set the starting z position of the zero-field-sub-world to be 1600cm
```

The average time per event consumed by `PHG4Reco` reduced from **161ms** to **67ms**, the most significant contribution comes from the energy threshold cut. The momentum resolution plot below shows there is essentially no difference with these additional time-saving features.

### DIFF
--- a/packages/PHField/CMakeLists.txt
+++ b/packages/PHField/CMakeLists.txt
@@ -65,6 +65,7 @@ add_library(phfield SHARED
   ${PROJECT_SOURCE_DIR}/PHField2D.cc
   ${PROJECT_SOURCE_DIR}/PHField3DCylindrical.cc
   ${PROJECT_SOURCE_DIR}/PHField3DCartesian.cc
+  ${PROJECT_SOURCE_DIR}/SQField3DCartesian.cc
   ${PROJECT_SOURCE_DIR}/PHFieldRegionalConst.cc
   ${PROJECT_SOURCE_DIR}/PHFieldSeaQuest.cc
   PHFieldUtility_Dict.C
--- a/packages/PHField/PHField.h
+++ b/packages/PHField/PHField.h
@@ -1,5 +1,3 @@
-
-
 #ifndef __PHFIELD_H__
 #define __PHFIELD_H__
 
@@ -7,7 +5,7 @@
 
 // units of this class. To convert internal value to Geant4/CLHEP units for fast access
 //#include <CLHEP/Units/SystemOfUnits.h>
-#include <G4SystemOfUnits.hh>
+#include <Geant4/G4SystemOfUnits.hh>
 
 #include <vector>
 
--- a/packages/PHField/PHField3DCartesian.cc
+++ b/packages/PHField/PHField3DCartesian.cc
@@ -61,7 +61,7 @@ PHField3DCartesian::PHField3DCartesian(const string &fname, const float magfield
          << "\n      Magnetic field Module - Verbosity:"
          << "\n-----------------------------------------------------------\n";
 
-  bool root_input = (filename.find(".root") != string::npos);
+  bool root_input = false;
   if(root_input) {
 		// open file
 		TFile *rootinput = TFile::Open(filename.c_str());
@@ -148,6 +148,11 @@ PHField3DCartesian::PHField3DCartesian(const string &fname, const float magfield
   ystepsize = (ymax - ymin) / (yvals.size() - 1);
   zstepsize = (zmax - zmin) / (zvals.size() - 1);
 
+  std::cout << "  its demensions and ranges of the field: " << fieldmap.size() << std::endl;
+  std::cout << "    X: " << xvals.size() << " bins, " << xmin/cm << " cm -- " << xmax/cm << " cm." << std::endl;
+  std::cout << "    Y: " << xvals.size() << " bins, " << ymin/cm << " cm -- " << ymax/cm << " cm." << std::endl;
+  std::cout << "    Z: " << xvals.size() << " bins, " << zmin/cm << " cm -- " << zmax/cm << " cm." << std::endl;
+
 #ifdef _DEBUG_ON
   auto key = boost::make_tuple(0, 0, 0);
   auto magval = fieldmap.find(key);
@@ -221,6 +226,7 @@ void PHField3DCartesian::GetFieldValue(const double point[4], double *Bfield) co
       point[1] < ymin || point[1] > ymax ||
       point[2] < zmin || point[2] > zmax)
   {
+    //std::cout << " sdsadsadsadasdasdasdas" << std::endl;
     return;
   }
   set<double>::const_iterator it = xvals.lower_bound(x);
--- a/packages/PHField/PHField3DCartesian.cc
+++ b/packages/PHField/PHField3DCartesian.cc
@@ -61,7 +61,7 @@ PHField3DCartesian::PHField3DCartesian(const string &fname, const float magfield
          << "\n      Magnetic field Module - Verbosity:"
          << "\n-----------------------------------------------------------\n";
 
-  bool root_input = false;
+  bool root_input = (filename.find(".root") != string::npos);
   if(root_input) {
 		// open file
 		TFile *rootinput = TFile::Open(filename.c_str());
--- a/packages/PHField/PHField3DCartesian.h
+++ b/packages/PHField/PHField3DCartesian.h
@@ -21,6 +21,10 @@ class PHField3DCartesian : public PHField
   //! @param[out] Bfield  field value. In the case of magnetic field, the order is Bx, By, Bz in in Geant4/CLHEP units
   void GetFieldValue(const double Point[4], double *Bfield) const;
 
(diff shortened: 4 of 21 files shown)


abinashpun:
Dear Kun,

Thank you very much for the bug fix and improvement. I am able to build it successfully but I ran into a couple of problems while running the macro you mentioned. 
-I think one of them is due to inclusion of your anamodule ( I am working on building one).
-2nd is I got the following error about not finding Fmag.root.
----------------------------------------
_Error in <TFile::TFile>: file /e906/app/software/osg/software/e1039/resource/geometry/magnetic_fields/Fmag.root does not exist
SQField3DCartesian: cannot find the input ROOT field map, abort /e906/app/software/osg/software/e1039/resource/geometry/magnetic_fields/Fmag.root_
-------------------------------------

liuk:
> Dear Kun,
> Thank you very much for the bug fix and improvement. I am able to build it successfully but I ran into a couple of problems while running the macro you mentioned.
> 
> * I think one of them is due to inclusion of your anamodule ( I am working on building one).
> * 2nd is I got the following error about not finding Fmag.root.
> 
> ```
> Error in <TFile::TFile>: file /e906/app/software/osg/software/e1039/resource/geometry/magnetic_fields/Fmag.root does not exist
> SQField3DCartesian: cannot find the input ROOT field map, abort /e906/app/software/osg/software/e1039/resource/geometry/magnetic_fields/Fmag.root_
> ```

Hello Abi, you do not really need to run my analysis module, it's only intended for myself to make quick QA plots. The updated core software should work with the old Fun4Sim macro with your own analysis module.

As for the field map, they are not updated on the official resource area yet, you can keep using the old tab.Fmag and tab.Kmag since they are sill supported in the new field map class. If you are interested in giving the ROOT field map a try, they can be found here: https://github.com/E1039-Collaboration/e1039-resource/tree/master/resource/geometry/magnetic_fields

abinashpun:
> > Dear Kun,
> > Thank you very much for the bug fix and improvement. I am able to build it successfully but I ran into a couple of problems while running the macro you mentioned.
> > 
> > * I think one of them is due to inclusion of your anamodule ( I am working on building one).
> > * 2nd is I got the following error about not finding Fmag.root.
> > 
> > ```
> > Error in <TFile::TFile>: file /e906/app/software/osg/software/e1039/resource/geometry/magnetic_fields/Fmag.root does not exist
> > SQField3DCartesian: cannot find the input ROOT field map, abort /e906/app/software/osg/software/e1039/resource/geometry/magnetic_fields/Fmag.root_
> > ```
> 
> Hello Abi, you do not really need to run my analysis module, it's only intended for myself to make quick QA plots. The updated core software should work with the old Fun4Sim macro with your own analysis module.
> 
> As for the field map, they are not updated on the official resource area yet, you can keep using the old tab.Fmag and tab.Kmag since they are sill supported in the new field map class. If you are interested in giving the ROOT field map a try, they can be found here: https://github.com/E1039-Collaboration/e1039-resource/tree/master/resource/geometry/magnetic_fields

Thank you, Kun. I am able to run the Fun4Sim successfully.

knakano0524:
Sorry for my late response.  Both the two parts are certainly helpful.  Since Abi has already run the test, I just looked over the files changed, which seem fine.

I just wonder if we should set the energy threshold to non-zero **by default**.  Actually I had thought that it was already non-zero.  If we set it in macro, we will need an extra care to consistently update all macros.  As it is not urgent nor conclusive, you can decide it before merging this PR.

liuk:
> Sorry for my late response. Both the two parts are certainly helpful. Since Abi has already run the test, I just looked over the files changed, which seem fine.
> 
> I just wonder if we should set the energy threshold to non-zero **by default**. Actually I had thought that it was already non-zero. If we set it in macro, we will need an extra care to consistently update all macros. As it is not urgent nor conclusive, you can decide it before merging this PR.

Hello Kenichi, thanks for the review and comment. I think it makes very good sense especially that in the default E906 simulation (GMC) a similar cut is applied at 1MeV. So this cut has been extensively tested in the muon simulation in some sense. I quickly run the test again using 1MeV as the threshold value and the result (running time and momentum resolution) is essentially the same. I will push a commit to turn this feature on in `PHG4Reco` with threshold set at 0.5MeV  by default.

liuk:
Hello all, thank you both for the review. I will merge this change later today. Since this change contains a bug fix that impacts everyone, we should update the official repository ASAP. There should not be any impact to the user other than that the simulation is faster.

abinashpun:
Dear Kun, 
Ok, I will send out the email and proceed to update our software version as soon as you merge it. 
Thank you.

PR #86, opened 2020-07-18
Re-organization of the global constants

liuk:
This PR changes the way how the global constants are handled in the simulation/reconstruction. In the past most of the global constants (like the position of magnets, basic track quality cuts, etc.) are hard-coded as C macros and included whenever it's needed. It was inherited from the E906 ktracker and works efficiently. But it requires one to recompile the whole software whenever a constant needs to be changed. It's particularly troublesome to work on the grid since it takes quite some time for the change to be propagated to all the worker nodes. 

This PR addresses this issue by moving all the global constants to the singleton object reconConsts, which user will need to initialize at the beginning of the analysis macro. The JobOptsSvc which was introduced at E906 to partly replace the global constants, is merged into the recoConsts as well. The initialization of the constants can be done in 3 different ways, after the initialization, user could change any specific parameters by its name. 
  1. Simply create/get an instance of recoConsts by: `recoConsts* rc = recoConsts::instance();`, a default set of constants will be initialized during the first call. The names and values of the constants are all hard-coded in `recoConsts::init_defaults()`.
  2. Initialize the constants by run number, like `recoConsts::instance()->init(runNo);`. We should provide such functions similar to how we select geometry parameters by run number using DbSvc. It is not implemented at this time;
  3. Initialize the constants by a pre-defined parameter set name, like `recoConsts::instance()->init("cosmic");` to load the parameters for cosmic ray tracking. The parameters can either be hard-coded in the recoConsts class or by a text file in the resource area which can be parsed by the recoConsts class. It's not implemented yet;

Other than this re-organization, this PR does not introduce any real changes. I have tested the code change by running an updated version of the RecoDev/RecoE1039Sim.C, the output is identical to PR85.

This PR is intended for discussion for now. I would suggest we go over the changes and see if this is how we would like to configure the simulation/reconstruction in the future. Eventually we should merge this change together with a usable cosmic ray reconstruction parameter set which Zulkaida is working on right now.


knakano0524:
Thank you for the vast amount of changes.  I did read through the files changed, and all looks fine.  We discussed several points about the implementation a few weeks ago via e-mail and you seem to have already taken all into consideration.  I would ask you to merge this PR whenever you are ready.

liuk:
Thanks @knakano0524 for the review! I added the parameter set for the cosmic ray reconstruction and included a cosmic ray generator after the PR was started last week. Below are the results of the reconstruction on both MC and run 1788. The result seems pretty reasonable to me, so I will merge this PR momentarily. @abinashpun, since this is a major change about how people uses the reconstruction, it would be nice if you could give a more detailed introduction on Thursday before we deploy this change.

Simulated cosmic muon: number of hits, reduced chi^2, tx, ty, x0, y0
Snipaste_2020-07-27_09-57-42Snipaste_2020-07-27_10-00-58Snipaste_2020-07-27_10-01-50Snipaste_2020-07-27_10-00-27Snipaste_2020-07-27_10-06-31Snipaste_2020-07-27_10-07-01

Cosmic muon from run 1788: number of hits, reduced chi^2, tx, ty, x0, y0
Snipaste_2020-07-27_10-07-53Snipaste_2020-07-27_10-35-56Snipaste_2020-07-27_10-29-58Snipaste_2020-07-27_10-09-40Snipaste_2020-07-27_10-30-33Snipaste_2020-07-27_10-31-07

abinashpun:
Dear Kun, thank you for the changes. I will act accordingly.

liuk:
> Dear Kun, thank you for the changes. I will act accordingly.

Hi Abi, thanks. I updated the macros I used for both data and MC cosmic reconstruction here: https://github.com/E1039-Collaboration/e1039-analysis/tree/master/RecoDev

abinashpun:
Dear Kun, Thank you. The Reco* macro fails as it is because *.root files for the magnetic field are still not available in the resource area. As a temporary solution for me I modified the recoConsts area to read tab.* files and started working.

Also, the Fun4Sim macro still tries to find the JobOptsSvc class for initialization and fails.

PR #87, opened 2020-07-25
Enhance the truth matching capability of TruthNodeMaker and added truth-based vertexing

liuk:
This PR consists of two primary changes, and some minor technical changes. It should not break any existing code.

## Update the `TruthNodeMaker`

- Currently `TruthNodeMaker` reads the `HepMCGenMap` to create the MC event header `SQEvent`. It will throw an ABORTEVENT if such node is not found, I disabled this behavior so that it can partly work with other generators. We will need a more permanent solution in the future;
- Currently `TruthNodeMaker` will read the `PHG4INEVENT` node to get all the primary track/dimuon information, it will also read the `G4HIT_*` containers on station-1 and station-3 to get the information of truth position and momentum of the tracks. I changed the latter part to use digitized `SQHit` instead since `SQHit` also contains the truth information and is more accurate since it's projected to the detector plane;
- If reconstructed tracks are available, the hit list of the truth tracks and reconstructed tracks will be matched. If the number of common hits is above a certain threshold (default value is 75%), the index of the reconstructed track will be saved in the `SQTrack` as well. User will be able to correlate truth and reconstructed track by `SQTrack::get_rec_track_id()`.

Now to use `TruthNodeMaker`, one needs to add it after the SQReco since it will use the reconstructed information as well. `G4TruthSubsystem` is of course also needed. Typical usage is the following:
```
TruthNodeMaker* truthMaker = new TruthNodeMaker();
truthMaker->set_legacy_rec_container(true);       //select the container type for the reconstructed info, more description is below
truthMaker->set_matching_threshold(0.75);         //set the truth matching threshold
se->registerSubsystem(truthNodeMaker);
```

## Introduction of Truth-based vertexing `SQTruthVertexing`

While we are working on a real vertex fitting package that works with GenFit, I created a vertexing package based on the truth vertex information so that we can use '**_reconstructed_**' dimuon information for various studies.  This package will read the truth track and dimuon information created by `TruthNodeMaker`, smear the truth vertex of single muon tracks and dimuons if enabled by the user, and calculate track/dimuon information at that vertex. It will need to be registered after `TruthNodeMaker`, typical usage is like the following:

```
SQTruthVertexing* truthvtx = new SQTruthVertexing();
truthvtx->set_legacy_rec_container(false);    // set the rec info container type like mentioned above, default is true
truthvtx->set_vtx_smearing(50.);                   // smear the truth z_vertex to mimic resolution effect, default is 0.
se->registerSubsystem(truthvtx);
```

## Introduction of fun4all-style container of reconstructed tracks and dimuons

I introduced two container classes `SRecTrackVector` and `SRecDimuonVector`, they contain the reconstructed tracks and dimuons and can be directly registered on the DST tree, to better conform to the overall fun4all style. They are intended to replace the original `SRecTrack` which contains a lot of duplicated information. Using `SRecEvent` as the container is still supported and is still the default setting, but users can and are encouraged to change to the new container by calling `set_legacy_rec_container(false);` in all the modules that deal with the reconstructed information. So far the lists include: `TruthNodeMaker`, `SQReco` and `SQTruthVertexing`.

## Minor bug fixes

- Abi found a bug about the truth position in `SQHit` that the z position is not a fixed value. It's now fixed;
- I removed several deprecated classes that are not being used any more;

## Simple tests

I generated 1.5k single muon events to check the momentum reconstruction at the vertex, shown below:
Screenshot from 2020-07-24 22-54-14
I also generated 10k DY events using pythia to look at the mass reconstruction at the vertex shown below. The reconstructed mass is shifted to high mass as we expected, asymmetric correction for the multiple scattering is needed to move the peak back to zero.
Screenshot from 2020-07-24 22-48-07


knakano0524:
`TruthNodeMaker` has lots of rooms for improvement.  Thank you for working on it.

About the info from `HepMCGenMap`, I remember Abi was trying to fetch the same info from other generators.  But it should be fine for now that ABORTEVENT is not thrown.

If we use `SQHIt` instead of `G4HIT_*`, I wonder what happens when we introduce the inefficiency and resolution into SQHit.  For example, if we just delete an inefficient SQ hit, the source info becomes unavailable.  There should be a way to cope with it, but do you have a rough prospect?

About `SRecTrackVector` (and also `SRecDimuonVector`), I imagine a cleaner way is to unify `SRecTrack` and `SQTrack` (by making a common base class) so that `SQTrackVector` can store both.  We could try it in a future.

liuk:
Hello Kenichi, thanks for the comments. Please see my answers below.

> `TruthNodeMaker` has lots of rooms for improvement. Thank you for working on it.
> 
> About the info from `HepMCGenMap`, I remember Abi was trying to fetch the same info from other generators. But it should be fine for now that ABORTEVENT is not thrown.
> 
This is what I meant. The track/dimuon info extracted from the `PHG4INEVENT` is very useful so it would be nice to let that part work first while we adapt the existing generators.

> If we use `SQHIt` instead of `G4HIT_*`, I wonder what happens when we introduce the inefficiency and resolution into SQHit. For example, if we just delete an inefficient SQ hit, the source info becomes unavailable. There should be a way to cope with it, but do you have a rough prospect?
> 
That's a very good point I didn't consider when making this change. I have updated the `TruthNodeMaker` so that if the `SQHit` is missing, it will fall back to the `G4HIT_` node to get the truth information by matching the `track_id`.

> About `SRecTrackVector` (and also `SRecDimuonVector`), I imagine a cleaner way is to unify `SRecTrack` and `SQTrack` (by making a common base class) so that `SQTrackVector` can store both. We could try it in a future.
>
This is a very interesting idea. I gave it some thought and this is what I came up with. Take `SRecTrack` for example, this class has about 100 methods used to set/get information. If we add all of them as pure virtual functions in the base class it probably looks very confusing and cumbersome. So I picked a small fraction of the functions from `SRecTrack` to `SQTrack`, one needs to explicitly do a dynamic_cast to access all functions in `SRecTrack`. Please take a look at the changes in commit [e1f9d4](https://github.com/E1039-Collaboration/e1039-core/pull/87/commits/e1f9d4b2d59f9a3fc46a8c7f56c0f027fbaf1c90) to see if you agree with this approach. If it makes sense I will make similar changes to SRecDimuon as well.

knakano0524:
By the way I found a bug in [line 144 of TruthNodeMaker.cc](https://github.com/liuk/e1039-core/blob/truthvtx/simulation/g4dst/TruthNodeMaker.cc#L144).  There `trk1` has to be `trk2`.  Could you fix it in this PR?

liuk:
> By the way I found a bug in [line 144 of TruthNodeMaker.cc](https://github.com/liuk/e1039-core/blob/truthvtx/simulation/g4dst/TruthNodeMaker.cc#L144). There `trk1` has to be `trk2`. Could you fix it in this PR?

Thanks for the careful check! The fix is pushed to the PR now.

liuk:
Hello guys, if there is no further comments/objections. I suggest we move forward with merging this PR by tomorrow.

knakano0524:
I'm sorry that I had forgotten to respond.  Thank you for the further updates.  The latest version looks OK.

PR #88, opened 2020-08-06
Implementation of the PHENIX EMCal 

liuk:
This PR implements the PHENIX EMCal by adding the following features:

- A SQCalHit is added to store the EMCal hit information;
- A SQEMCalSubsystem and its corresponding detector/stepping functions have been implemented;
- A digitizeEMCal function has been added to the digitizer to convert g4hits to SQCalHit

Since this PR does not touch any of the existing behavior. I expect it brings no change to the primary user. I run a simple 1k single muon generation using the code from the current master and this PR, the results are indeed identical.


knakano0524:
Sorry for my late comment.  The implementation looks great.  Just a minor question is that
[the name of `CellType` in SQHit.h](https://github.com/E1039-Collaboration/e1039-core/pull/88/files#diff-a3d8096eab9cce8bb32bc97f61ce783cR22) is a bit confusing to me.  It sounds like a detector type like PbGl/PbSc but is actually a set of energy deposits.  Would you conceive a better name such as `CellEnergyList` and `EnergyDepositMap`?

liuk:
Thanks Kenichi, for the suggestion. It's indeed very confusing. I followed the example code from [sPHENIX](https://github.com/sPHENIX-Collaboration/coresoftware/blob/e91d61170db2f5c9e0555dd2fb795107a63a1693/offline/packages/CaloBase/RawTower.h#L21) and renamed it `CellMap`. 

In addition to that, I also re-organize the hit type implementation for the EMCal and modified/created two hit classes: SQCalHit_v1 and SQCalMCHit_v1 basically following the structure of SQHit_v1 and SQMCHit_v1. I hope this make things clearer.

knakano0524:
Thank you for the response, Kun.  The updates look all fine.  I would ask you to merge this PR whenever you feel comfortable.

PR #90, opened 2020-09-10
Translation from E906 SRawEvent to E1039 DST-style data structure

liuk:
In the past Haiwang implemented a Fun4AllSRawEventInputManager to read the E906-style ROOT data file which contains the SRawEvent, and add it to the DST tree. 

I added an option `Fun4AllSRawEventInputManager::enable_E1039_translation()` which is turned off by default. User can call this function to turn it on, then in addition to adding a `SRawEvent` node to the DST tree, the input manager will also use the information from `SRawEvent` to create `SQRun` and `SQSpillMap` node on the RUN tree, plus `SQEvent`, `SQHitVector`, `SQTriggerHitVector` on the DST tree.  The main purpose is to help people make analysis modules, in particular detector analysis modules for both E1039 and E906 in the same way.

Since this is turned off be default and does not alter the original behavior, it should have very little if not no effect on the end user.

### DIFF
--- a/packages/reco/ktracker/Fun4AllSRawEventInputManager.cxx
+++ b/packages/reco/ktracker/Fun4AllSRawEventInputManager.cxx
@@ -122,8 +122,6 @@ void Fun4AllSRawEventInputManager::enable_E1039_translation()
 
 void Fun4AllSRawEventInputManager::E906ToE1039()
 {
-  if(!_enable_e1039_translation) return;
-
   run_header->set_run_id(_srawEvent->getRunID());
 
   SQSpill* spill = spill_map->get(_srawEvent->getSpillID());
@@ -133,7 +131,7 @@ void Fun4AllSRawEventInputManager::E906ToE1039()
     spill->set_run_id(_srawEvent->getRunID());
 
     spill_map->insert(spill);
-    run_header->set_run_id(run_header->get_run_id()+1);
+    run_header->set_n_spill(spill_map->size());
   }
 
   event_header->set_run_id(_srawEvent->getRunID());
--- a/packages/reco/ktracker/Fun4AllSRawEventInputManager.cxx
+++ b/packages/reco/ktracker/Fun4AllSRawEventInputManager.cxx
@@ -2,16 +2,15 @@
 
 #include "SRawEvent.h"
 
-//#include <event/EVIO_Event.h>
-//#include <interface_main/SQHit_v1.h>
-//#include <interface_main/SQHitVector_v1.h>
-//#include <interface_main/SQEvent_v1.h>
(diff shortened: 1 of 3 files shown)


knakano0524 (commented):
I'm sorry for my late response.  Thank you for working on this data translation.  I looked over the changes.  I think they are overall fine and just made a few code-wise comments inline.

Thinking of its usage, do we expect that this input manager will be the standard way of reading the E906 data in E1039?  If so, the E1039 decoder (i.e. `online/decoder_maindaq/Fun4AllEVIOInputManager`) should be able to handle only E1039 runs and the E906 decoder should be kept maintained by E906 and E1039.

One remark I see is that we will have to keep the SRawEvent class in E1039 and that in E906 identical.  I think it is feasible (assuming that SQEvent etc. will be the main data container in E1039 and  SRawEvent will be used only as interface), but we should be careful about keeping the identity.

liuk:
> I'm sorry for my late response. Thank you for working on this data translation. I looked over the changes. I think they are overall fine and just made a few code-wise comments inline.
> 
> Thinking of its usage, do we expect that this input manager will be the standard way of reading the E906 data in E1039? If so, the E1039 decoder (i.e. `online/decoder_maindaq/Fun4AllEVIOInputManager`) should be able to handle only E1039 runs and the E906 decoder should be kept maintained by E906 and E1039.
> 
> One remark I see is that we will have to keep the SRawEvent class in E1039 and that in E906 identical. I think it is feasible (assuming that SQEvent etc. will be the main data container in E1039 and SRawEvent will be used only as interface), but we should be careful about keeping the identity.

Thanks a lot for the careful checks. The two places you spotted are obviously my bugs. I have fixed them in the new commit.

`Fun4AllSRawEventInputManager` was introduced by Haiwang so that we can run kTracker with the E906 data under the fun4all framework, by reading the `SRawEvent` and putting it on the DST tree for ktracker to use. On top of Haiwang's work, this change is introduced to optionally convert the `SRawEvent` to `SQ*` data format so that other detector performance analysis packages developed for the E1039 could be used on the existing E906 data as well. I believe we can still update the `SRawEvent` as we need, but just need to be careful to increment the ROOT dictionary version number whenever we make changes to the data members.

In summary, `Fun4AllSRawEventInputManager` is intended to provide an interface to access the existing E906 data that has already been decoded.

knakano0524:
Thank you for the quick fix.  I think the code is ready to be merged.  Would you please do that whenever you like?

About the contents of `SRawEvent`, I think every version of concrete classes has to be kept in the source (like `SQSpill_v1` and `SQSpill_v2` in `interface_main/`) in order to make every version readable in the Fun4All framework, although I have never tried it.  Anyway we can discuss such implementation just when we need it.

PR #91, opened 2020-10-09
Small corrections on trigger emulator (DPTriggerAnalyzer)

knakano0524:
Here is a small update on the trigger emulator, `DPTriggerAnalyzer`.

The most important correction is to make this module **not** resetting run, spill and event IDs of SQEvent at the beginning of process_event().

Another important(?) correction is to stop using `GeomSvc::getTriggerLv()`.  It sounds to have a different purpose (as the trigger level usually means L0/1/2 of V1495 boards now) and its code depends on the order of detector IDs.  Instead I introduced `GeomSvc::getHodoStation()`.

Other corrections are just to clean up the code;
 - Removed all hard-coded detector IDs and
 - Removed lots of unused functions and variables.

The trigger logic emulated by DPTriggerAnalyzer is right, as far as I looked into the code.  I confirmed that the emulated results are identical before vs after my change, using my J/psi in-acceptance events.


packages/geom_svc/GeomSvc.cxx
```diff
@@ -206,7 +206,7 @@ std::ostream& operator << (std::ostream& os, const Plane& plane)
 }
 
 GeomSvc* GeomSvc::p_geometrySvc = NULL;
-bool     GeomSvc::use_dbsvc = false;
+bool     GeomSvc::use_dbsvc = true;
```
knakano0524, line 209: I forgot to mention one important change.  I would propose to make GeomSvc use DbSvc by default.  All macros will no longer need to call `GeomSvc::UserDbSvc(true)`.

knakano0524 (commented):
Additional explanation.

abinashpun:
I built it successfully. Also, compared the dimuon outputs (from PYTHIA) with pr90. The results are identical. As I am not familiar with the process in details, I have not looked at the codes carefully. Please let me know if you want me to check anything specifically.

knakano0524:
Thank you for checking the dimuon outputs and finding the bug.  I have fixed it.  I think our checks are already sufficient.  I will merge this branch in the coming weekend.

knakano0524:
Forhad asked me why NIM1 and NIM2 are not imitating MATRIX1 and MATRIX2 in the current `DPTriggerAnalyzer` module.  It is because the logic emulated now is based on the E906 configuration.  

Actually I don't remember how the E1039 logic is at present.  If it is provided, I can implement it.  But it won't be very simple for me to implement multiple (run-dependent) types of the logic because the NIM and FPGA L2 parts are hard-coded in `DPTriggerAnalyzer`.  If we want such capability, we should first discuss how to record the trigger logic per run.

PR #92, opened 2020-11-06
Improvements for study of trigger acceptance

knakano0524:
This change is aimed at improving various functions needed/wanted for the study of the trigger acceptance.

Here is a summary of the main updates;
- `DPTriggerAnalyzer` is now able to change its NIM emulation mode (AND vs OR) via `set_nim_mode()`.
- `TruthNodeMaker` now creates the `SQEvent` node in the DST branch, which had been created by `DPTriggerAnalyzer` but prematurely omitted by me in PR 91.
- `SQPrimaryParticleGen` and related classes were modified in order to sort out its output data structure and fix a bug.

Regarding the last item, 
- I think to switch its output data container from `SQDimuonTruthInfoContainer` to `SQDimuonVector`.  Both the containers are available at present, and I have confirmed that their contents are identical (as analyzed with `e1039-analysis/AnaSimDst`).
- I found that the cross section (= event weight) given by `SQPrimaryParticleGen` is sometimes (at 0.2%) NaN.  It was caused by a tiny bug in `SQPrimaryVertexGen::traverse()`, namely `pARatio` (=`protonPerc`) became incorrect when  `A` and `Z` of `SQBeamlineObject` was modified.  Particularly for the air gaps at line 209, pARatio = 0/0 = NaN.  I fixed it by changing "SQBeamlineObject::protonPerc" from variable to function.

I would like to discuss and decide the following things before closing this pull request.
1. Can we remove `SQMCDimuon`?  Its key function is `calcVariables()`, which has been exported to `UtilDimuon::CalcVar()`.
1. Can we remove (or comment out) the PYTHIA mode of `SQPrimaryParticleGen`?  It is not fully functional and the standard generator (`PHPythia8`) already exists.


abinashpun:
Dear Kenichi,

Thank you so much for the fixing the bug in event weight, tidying up the legacy generator code and other additions. 

1. Can we remove `SQMCDimuon`? Its key function is calcVariables(), which has been exported to `UtilDimuon::CalcVar()`.
Yes, I think we can remove it. Also, I think we can add the features in the `SQDimuonTruthInfoContainer` to the `SQDimuon` and get rid of it too.

1. Can we remove (or comment out) the PYTHIA mode of `SQPrimaryParticleGen`? It is not fully functional and the standard generator (PHPythia8) already exists.
I think the PYTHIA mode in the `SQPrimaryParticleGen` has some nice features. For eg. it uses the legacy vertex generator and chooses the p-n vs. n-n interaction based on the p/A ratio of the vertex material. However, it lacks some other features like trigger condition and also somehow the configuration file can't be passed to it. I tried to work on it when I first committed it but didn't get time to look back and fix it. So, depending on how important those features you think, we can either develop it or comment it out for now.

knakano0524:
Thank you for the reply.  I have removed `SQDimuonTruthInfoContainer` and `SQMCDimuon`.  The output values (like mass, x1, etc.) stay the same down to 1e-5 at least.  I left the PYTHIA mode of `SQPrimaryParticleGen` unchanged.

I want to merge this PR at the end of tomorrow, Friday.  If you would have further questions/comments, please let me know by that time.

abinashpun:
Dear Kenichi, 
Thank you for the changes. I have no further comments.

PR #93, opened 2020-11-06
Small change of OnlMonHodo tdcTime range

knakano0524:
The time axis of the OnlMonHodo plots is now auto-adjusted so as to closely show all recorded hits.  This change was asked by the hodo experts, since the signal timing might change frequently in future trigger timing adjustments.  The new code is already running fine for the online decoding.

The new option (-b) of `build.sh` might be helpful to others, which let us put the `build` directory on a faster disk.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,7 +9,8 @@
 #   ./build.sh [-c cmake_args] Optionally pass addtional cmake args to the build
 #
 # The 2nd usage is not recommended but kept available for backward compatibility for now.
-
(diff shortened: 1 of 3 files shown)


abinashpun:
Dear Kenichi,
I just did the sanity check of building it. It works fine.

knakano0524:
Dear Abi,
Thank you so much for the sanity check.  I believe we can merge this now.

PR #95, opened 2020-12-08
Performance optimization for the tracking-related code

liuk:
This change does not introduce any real change, but in general some minor performance twist and code re-structure.
* in GeomSvc  unordered_map instead of map to store the wire position lookup table to achieve o(1) performance
* added more information in the G4Hit useful for EMCal study
* make all member functions const whenever possible in GFMeasurement and Tracklet class
* make GFFitter::processOneTrack return different exit code when encountered different failures to assist debugging

I have tested the change using Fun4Sim program, before and after the change yielded exactly the same output, as expected.

Obviously this PR is not in a hurry, and can be rolled out with the next big update.


knakano0524:
Sorry for my late reply.  I looked through all the changes.  All are fine.  I would ask Kun (or Abi) to merge this PR when appropriate.

PR #97, opened 2021-01-11
Updates about handling of E906 data and E1039 online data

knakano0524:
This request includes multiple updates about handling of the E906 data and also the E1039 online data.  The updates should not affect the general behavior of `e1039-core` because they are localized in SubsysReco module and Fun4All macro.  I have confirmed that they are working as intended.  Thus I think it sufficient if you could just check the concept/validity of the changes, rather than testing them one-by-one.

# Handling of E906 data

I made several changes/fixes that were needed for my analysis on the trigger background rate via the trigger emulation.

* DPTriggerAnalyzer
    - Can use the trigger hit (instead of the normal hodo. hit) when emulating the trigger logic.
    - Can require the in-time cut.
    - Can create a new node (`SQEvent`) to store the emulated result, instead of overwriting the existing node.
* Fun4AllSRawEventInputManager
    - A bug was fixed, which filled the normal hit into the trigger-hit list (`SQHitVector`).

# Handling of E1039 online data

I made three changes in accordance with problems found on the online data decoding/monitoring.

* An online-monitoring parameter (`OnlMonComm::m_n_sp_sel_max`) was decreased, which sets the max number of spills per run where the spill-by-spill OnlMon histograms are kept.  It was found so large as to use up the server memory in case of long runs.
* A method of recording TDC board problems in the online decoding (`DecoError.cc`) was modified.  It caused a (rare) error when re-creating a database table.  It now doesn't recreate it but keeps using the existing one.
* `EventDisp4MainDaqDst.C` (and related files) was modified, which is the macro to run the online event display.  The usage of the geometry configuration macros (e.g. `G4_Target.C`) in it was not up-to-date.


liuk:
Hello Kenichi, many thanks for the changes. I have reviewed the offline part and find no problems. Thanks for spotting the bug in the Fun4SRawEventInputManager. Please go ahead and merge the PR.

knakano0524:
Thank you for the review, Kun. OK, I will merge the PR in a few hours.

PR #98, opened 2021-02-12
DSTs per spill group and longer MySQL connection timeout

knakano0524:
Two small updates are included.  (1) `Fun4AllSpillDstOutputManager`, which outputs one DST file per spill group, is enabled in the online decoder.  The number of spills per DST is set to 100 now.  (2) The connection timeout was made longer (from 10 to 120 sec) in order to hopefully resolve the error in DbSvc::ConnectServer().

I cannot assure that the DbSvc error has been really resolved because the error was rare (0 in 10,000 trials) even without this update.  I expect to check it in the coming practical uses (simulation, decoding, etc.).

The updates are simple, and I have confirmed that the updated code can be compiled and run fine.  Thus I think Abi can just take a look and merge.


abinashpun:
Dear Kenichi,

Thank you for the changes. I will merge this pull request.

PR #100, opened 2021-03-20
Addition of utility functions.

knakano0524:
Here I am adding classes/functions to`packages/UtilAna/`.  For example
* `UtilDimuon::FindDimuonByID()` finds a dimuon by its ID (not array index) from `SQDimuonVector`.
* `UtilTrack::FindHitsOfTrack()` finds all hits that are associated with a track.

If you have an idea of any other useful functions, I can add such this time.  Since the new classes/functions do not change the default behavior of `e1039-core`, I expect you could merge this PR after taking a brief look at the changes.

### DIFF
--- a/packages/UtilAna/LinkDef.h
+++ b/packages/UtilAna/LinkDef.h
@@ -5,6 +5,7 @@
 
 #pragma link C++ class UtilOnline-!;
 #pragma link C++ class UtilSQHit-!;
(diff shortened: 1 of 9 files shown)


abinashpun:
Dear Kenichi,

Thank you for adding the interfaces. I have briefly gone through the changes and did the sanity check by building and running it successfully. I will merge them try to use them in my analysis.

PR #101, opened 2021-05-05
Addition of OnlMon client for QIE.

knakano0524:
Here I added a new OnlMon client for QIE since the QIE TDC board was mounted on ROC and started being read out last week. The new version is already running online and will output contents in OnlMon plots once the board starts sending out data, as Zong-Wei reported in DocDB 9277.

I expect this pull request could be merged just after you shortly review the changes, because the new version is already running and does not affect other online/offline uses.

### DIFF
--- a/online/decoder_maindaq/MainDaqParser.cc
+++ b/online/decoder_maindaq/MainDaqParser.cc
@@ -805,95 +805,93 @@ int MainDaqParser::ProcessBoardTriggerCount (int* words, int j)
  */
 int MainDaqParser::ProcessBoardFeeQIE (int* words, int idx)
 {
-    if (dec_par.runID < 22400) Abort("feeQIE does not support run < 22400.");
+  while (words[idx] == (int)0xe906e906) idx++; // Still necessary??
 
-    while (words[idx] == (int)0xe906e906) idx++; // Still necessary??
-
-    // int boardID = get_hex_bits (words[idx], 7, 4);
-    int n_wd = get_hex_bits (words[idx], 3, 4);
-    int idx_end = idx + n_wd + 1; // It points to the 1st word of the next board.
-    idx++;
-    if (n_wd == 0) return idx;
-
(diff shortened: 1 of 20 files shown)


online/decoder_maindaq/MainDaqParser.cc
```diff
@@ -805,95 +805,93 @@ int MainDaqParser::ProcessBoardTriggerCount (int* words, int j)
  */
 int MainDaqParser::ProcessBoardFeeQIE (int* words, int idx)
 {
-    if (dec_par.runID < 22400) Abort("feeQIE does not support run < 22400.");
```
knakano0524, line 808: The only meaningful change in this file is the removal of this line.  This run-number condition is valid only for E906 data.

The other changes in this file are just of the size of the line indent.

online/onlmonserver/OnlMonCham.cc
```diff
@@ -121,20 +121,20 @@ int OnlMonCham::DrawMonitor()
 {
   OnlMonCanvas* can0 = GetCanvas(0);
   TPad* pad0 = can0->GetMainPad();
-  pad0->SetGrid();
```
knakano0524, line 124: Here is an extra change that I made in this PR.  The axis grid of all OnlMon plots is (correctly) drawn now.  You can look at the OnlMon HTML output of run 2868, for example, to see how they appear.

packages/UtilAna/UtilBeam.h
```diff
@@ -0,0 +1,9 @@
+#ifndef _UTIL_BEAM__H_
+#define _UTIL_BEAM__H_
+
+namespace UtilBeam {
```
knakano0524, line 4: I added this namespace to collect useful functions about the beam.

liuk:
Hello Kenichi, thanks for the changes. I have briefly gone over the changes and find no problems. Please go ahead with the merge.

knakano0524:
Hello Kun, thanks for the prompt check.  OK, I'm going to merge now.

PR #103, opened 2021-06-02
Isolating vertex generator 

abinashpun:
This pull request is intended for isolating the legacy vertex generator so that it can be used for other generators. I have separated the module for vertex generation in separate repository "E906LegacyVtxGen". 

So, I have made some modifications from the Kun's code. The major changes (by Kun and me) compared to current codes are listed as follows;

- Using TGeoManager to automatically find the beam line volumes 
- Using realistic beam profile (double gaussian core and 1/r tail: Thanks to Kenichi)
- Providing interface to PHG4ParticleGun, PHG4SimpleEventGenerator 
- Providing interface to PHHepMCGenHelper. So, any generators using PHHepMCGenHelper to set the vertex (including PHPythia8) can use legacy vertex generator.
- Using configuration file for the pythia in legacy generator

In this pull request, I have not included SQExternalGen from Kun. I will do more test about it and give another pull request for that. Following are some plots with gun generator using the legacy vertex generator and beam profile
image
image
image


abinashpun:
I have committed following changes;
1. fNpx and fNpy changed to 200x200 from 100x100 (for beam profile function)
2. added interfaces for using target only or fmag only. You can turn on flag by `SQPrimaryVertexGen::set_targetOnlyMode();` and `SQPrimaryVertexGen::set_dumpOnlyMode();` in your macro.

Target only mode:
image

FMAG only:
image

generators/E906LegacyVtxGen/SQPrimaryVertexGen.h
```diff
@@ -0,0 +1,106 @@
...
+  void calcProb();
+
+ public:
+  unsigned int nPieces;
+  double probSum;
+  double accumulatedProbs[100];
+  std::vector<SQBeamlineObject> interactables;
+};
+
+class SQPrimaryVertexGen: public SubsysReco
```
knakano0524, line 37: Do we still want this inheritance?  Will `SQPrimaryVertexGen` be registered to Fun4AllServer in some use cases?

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
```diff
@@ -0,0 +1,270 @@
...
+  
+  const double r_pipe = 5.0; // in cm
+  const double r_boun = 3.0; // normalized boundary, where gaus changes to 1/r
+  if (pow(x,2) + pow(y,2) > pow(r_pipe,2)) { // not "x-x_c" nor "y-y_c" here
+    return 0.0;
+  }
+
+  //@ for target only mode
+  recoConsts* rc = recoConsts::instance();   
+  bool inTarget = (pow(x-rc->get_DoubleFlag("X0_TARGET"),2)/pow(rc->get_DoubleFlag("RX_TARGET"),2 )+ pow(y-rc->get_DoubleFlag("Y0_TARGET"),2)/pow(rc->get_DoubleFlag("RY_TARGET"),2)) < 1.;
```
knakano0524, line 252: This part seems unnatural.  In physics, the beam profile should not depend on any condition of target/dump materials.  In technique, this part makes the `_targetOnlyMode` variable static.

Why do you need this part?  I expect the target-only mode can be fulfilled by the target side, i.e. `fillMaterialProfile()` only.

simulation/g4main/PHG4ParticleGun.cc
```diff
@@ -52,16 +51,18 @@ PHG4ParticleGun::process_event(PHCompositeNode *topNode)
...
+		
   PHG4InEvent *ineve = findNode::getClass<PHG4InEvent>(topNode,"PHG4INEVENT");
   ReuseExistingVertex(topNode); // checks if we should reuse existing vertex
 
-
-  // cout<<" Here I am in particle gun...."<<endl;
-  //cout << "PHG4ParticleGun: {" << vx << ", " << vy << ", "<<vtx_z<< "}" << endl;
+  //! using vertex generated by E906 legacy generator
+  if ( _legacy_vertexgenerator) 
+	{TVector3 vtx        = _vertexGen->generateVertex();
```
knakano0524, line 60: Better make a new line after `{`.

knakano0524 (commented):
Thank you for sorting out the functions of our code.  I went through the changes and had the following comments.  I haven't run the updated code.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
abinashpun: Yes. But if we choose target only condition in fillMaterialProfile(), it will cause crash because of not finding any material for vertex (x,y) outside the target for the event. Also, this condition in beam profile will only generate (X,Y) from the part where the target can interact without having to modify the actual gaussian function (inner core and outer tail). 

But for FMAG, this won't happen since the beam profile we defined falls inside the range of the width of FMAG.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.h
abinashpun: I tried to make it a subsysreco class initially but later did not really needed to register to Fun4All. In principle, it should work out without this inheritance.  But somehow, the macro is crashing when I removed it by giving the following error message;
"Fun4AllServer::setRun(): could not get timestamp for run  0, using tics(0) "
I will look into the details and come back to it.

knakano0524: That error message always appears at present.  So the crash should be caused by somewhere else.  It's great if you could remove this inheritance, in order to avoid misuse of this class.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
knakano0524: That sounds a problem specific to `fillMaterialProfile()`.  It should check and properly handle the case of no material for any geometry configuration.  Can you add such a check?

generators/E906LegacyVtxGen/SQPrimaryVertexGen.h
abinashpun: I will look into it thank you.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
abinashpun: I will try to look into it. Thank you Kenichi.

simulation/g4main/PHG4ParticleGun.cc
abinashpun: I will change it accordingly. Thank you.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
```diff
@@ -0,0 +1,270 @@
...
+TVector3 SQPrimaryVertexGen::generateVertex()
+{
+  if(!inited) init();
+
+  double xvtx, yvtx, zvtx;
+  generateVtxPerp(xvtx, yvtx);
+
+  bool inTarget = ((xvtx-targetX0)*(xvtx-targetX0)/targetSX/targetSX + (yvtx-targetY0)*(yvtx-targetX0)/targetSY/targetSY) < 1.;
+  MaterialProfile* activeProf;
+  if(!inTarget)
```
knakano0524, line 207: What is the purpose of this "if" condition?  I understand that the material profile at x = y = 0 (i.e. `defaultMatProf`) is used when the beam position generated is inside the target area.  But I wonder why such a special handling is necessary.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
```diff
@@ -0,0 +1,270 @@
...
+    {
+      
+      activeProf = new MaterialProfile;
+      fillMaterialProfile(activeProf, xvtx, yvtx);
+    }
+  else
+    {
+      activeProf = defaultMatProf;
```
knakano0524, line 215: A real issue here is that `nPieces` is not updated in this "else" case, since `fillMaterialProfile` is not called.

I think `nPieces` should not be a member variable and instead `interactables.size()` should be called.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+  int index = activeProf->findInteractingPiece(rndm.Rndm());
```
knakano0524, line 218: This line (or next) gets into trouble when the number of interacting pieces is zero.  Here we should repeat the generation of xvtx and yvtx until the number becomes non-zero.

knakano0524 (commented):
I looked into the handling of interacting materials.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.h
abinashpun: The fresh build worked. Now it is not a SubsysReco class

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
abinashpun: I assume @liuk implemented this condition to save time to fill the material profile inside the target area as we don't expect to change any geometry in that region. In principle, it should work even if each time we fill the material profile corresponding to (x, y) values. But @liuk  might can give the exact answer for this.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
abinashpun: Thank you Kenichi for careful look. Here, for the else case, the "nPieces" should be constant as the fillMaterialProfile() is called initially for the defaultMatProf in init() function.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
abinashpun: Thank you Kenichi for the suggestion. I have deleted the portion inside the "funcBeamProfile" and added condition for generating the xvtx and yvtx inside the target area for _targetOnlyMode. I have pushed the commits with your suggestions.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
knakano0524: If we keep this, could you add a comment to the code that explains the purpose and the assumption (the material profile inside the target area must be uniform)?

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
knakano0524: The issue should happen when the "if" case is executed in an event and later the "else" case is executed in another event.  `nPieces` is changed in the "if" case (for the outside of the target area), and then it is not changed back in the "else" case (for the inside).

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
knakano0524: I would like to take a more-general method for this.  I understand that the target-only mode and other special modes are reflected in `MaterialProfile` by `fillMaterialProfile()`.  Thus `generateVertex()` need not worry about the mode (`_targetOnlyMode`) but generates the x,y vertex until `activeProf->interactables.size() > 0` in principle.

I'm afraid that the present method is so specific to cause a confusion/bug in future.  

Can you also make `_targetOnlyMode` and `_dumpOnlyMode` not static?

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
abinashpun: I see. Yes, you are right. I will look into it.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
abinashpun: I have changed `_targetOnlyMode` and `_dumpOnlyMode` to non-static one (I forgot why I set them like that in first place). Thank you for your pointing me that out.

Also, I will try to look into more-general method for this.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
liuk: I will give it a try to not inheriting from the SubsystemReco

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
liuk: Current implementation of target only and dump only mode is not very good since it introduces expensive string comparison many times during the generation process. I would use start/stop z position for the geometry navigator instead to limit the search for the interacables to target/dump only locations. 

Let me push a change to this PR today to address this issue.

generators/E906LegacyVtxGen/SQPrimaryVertexGen.C
abinashpun: Yes, now it is not inheriting from the SubsysReco. Thank you.

liuk:
Hello guys, I made a few changes to the PR, the detailed changes are summarized in commit message of 57822c6. I hope I addressed all the issues discussed here but please take a look to double-check. 

I have run the test locally on my computer and everything seems correct. Abi please give it a try using your test macro and see if it agrees with your previous results. Thanks.

liuk:
I submitted another change according to a discussion with Abi. Now the target/dump only mode can either be toggled directly, or set though recoConsts. I added two boolean flags TARGETONLY and DUMPONLY to the recoConsts class, user can set it before initialization the fun4all system. vertex generator will read these flags to toggle corresponding modes. Both flags are defaulted to be false.

test/test_vtx.C has been changed to reflect these changes as well.

knakano0524:
I would agree to merge this now.  Note that the handling of `nPieces` is fine as is, since I had mis-understood that it was a member variable of `SQPrimaryVertexGen` instead of `MaterialProfile`.

I have resolved the conflicts.  Could you check the code and build it?  I use the browser interface to resolve the conflicts this time.  I expected that you and I worked on it part-by-part, but I had to resolve everything to commit the change.

knakano0524:
I tried to build it and failed.  Line 293 of `SQPrimaryParticleGen.C` has to be fixed as follows.  Could you fix it and then merge this PR?

```
-int SQPrimaryParticleGen::generateDrellYan(TVector3 vtx, const double pARatio, double luminosity)
+int SQPrimaryParticleGen::generateDrellYan(const TVector3& vtx, const double pARatio, double luminosity)
```

abinashpun:
> I tried to build it and failed. Line 293 of `SQPrimaryParticleGen.C` has to be fixed as follows. Could you fix it and then merge this PR?
> 
> ```
> -int SQPrimaryParticleGen::generateDrellYan(TVector3 vtx, const double pARatio, double luminosity)
> +int SQPrimaryParticleGen::generateDrellYan(const TVector3& vtx, const double pARatio, double luminosity)
> ```

Thank you Kenichi. I also got the same error. I will fix it, test it and merge the pull request. I think we should update our default software after this merge. Please let me know what you (@liuk , @knakano0524 ) think.

PR #104, opened 2021-06-21
Switchable geometry info

knakano0524:
This PR makes the geometry info switchable.  I needed this switch when confirming that the track position dependence of the sagitta ratio observed by Dinupa was caused by the plane tilt angle.  It will be needed when we have multiple versions of the geometry info in the future.  The default geometry hasn't changed.

I first added three constants to `recoConsts`; `RUNNUMBER`, `DB_SERVER` and `DB_USER`.  I then modified `DbSvc`, which now uses `DB_SERVER` and `DB_USER` by default to select the server and the user of DB.  I also modified `GeomSvc` so that `initPlaneDbSvc()` now reads a run number from `RUNNUMBER` and initializes the plane position for the run (via `GeomParamPlane`).

The plane position per run range is given in `e1039-resource/geom/plane/run_range.tsv`.  It has not changed for the default run number (i.e. `G9_run5_2` for RUNNUMBER = 1).  An ideal plane position (which has no plane tilt angle) was newly created for runs 90000-99999.  Thus user can select the ideal geometry by adding the following lines to `Fun4Sim.C` for example;
```
  recoConsts *rc = recoConsts::instance();
  rc->init(90000); // dummy run number for zero-plane-angle geometry
  rc->set_BoolFlag("IdealGeom", true);
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -58,8 +58,8 @@ if [ $mode = 'single' ] || [ $mode = 'increment' ]; then
 else # 'all' or 'resume'
   declare -a packages=(
     packages/global_consts
-    packages/db_svc
     framework/phool
+    packages/db_svc
     packages/geom_svc
     framework/ffaobjects
     framework/fun4all
--- a/framework/phool/recoConsts.cc
+++ b/framework/phool/recoConsts.cc
@@ -157,6 +157,7 @@ void recoConsts::set_defaults()
 void recoConsts::init(int runNo, bool verbose)
 {
   //TODO: initialization based on run range
+  set_IntFlag("RUNNUMBER", runNo);
   return;
 }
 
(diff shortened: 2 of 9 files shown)


abinashpun:
Dear Kenichi,

Thank you for the changes. I did a sanity check by building it successfully. Also, roughly have a look at the changes. This PR has addressed the concern from @Hugo-Leung  that the SQDB pwd being public in the github code. Unless, Kun has some comments, I think we can merge this PR.

Abi

knakano0524:
Thank you again for the checks.  I'd like to merge this PR and also #105 at noon tomorrow.  If you have further comments, please let me know by that time.

It is surely better that the new code doesn't contain pwd.  But all the past versions are still available on GitHub and thus ideally we should change pwd in a future, probably when we select the final DB server for E1039.

abinashpun:
> Thank you again for the checks. I'd like to merge this PR and also #105 at noon tomorrow. If you have further comments, please let me know by that time.
> 
> It is surely better that the new code doesn't contain pwd. But all the past versions are still available on GitHub and thus ideally we should change pwd in a future, probably when we select the final DB server for E1039.

Dear Kenichi, I don't have any further comments on the pull requests.

PR #105, opened 2021-06-22
Simulation constants and chamber realization

knakano0524:
Two topics are included in this PR.

1. Management of simulation constants.  
I'm thinking a method of properly managing all simulation parameters, as presented in DocDB 9331.  The method proposed in this PR is to simply make use of `recoConsts`.  As a demonstration, I put several lines in `PHPythia8` etc. that add parameters to `recoConsts`.  The contents of `recoConsts` are expected to be written to a TSV file at the end of `Fun4Sim.C` etc. via `rc->WriteToFile("recoConsts.tsv")`.  If you are fine with this method, I will be adding more lines in a future.

2. Implementation of chamber realization.
I made a new SubsysReco module, `SQChamberRealization`.  Here I would propose to have this module outside `SQDigitizer`, because the chamber realization is sometimes re-done with different parameters.  I added the drift-time resolution (`dt`) to the X-T parameter set (which is stored in `CalibParamXT`), since the realization needed it when computing the smeared drift time from the drift distance.  I also modified `PHG4TruthSubsystem` in order to create the SQRun node and set a run number in it, since the realization should depend on run number.  Note that the current X-T parameter set is just a copy of the last E906 one.

I confirmed that a low chamber efficiency results in a low track reconstruction rate quantitatively, using `e1039-analysis/SimChainDev/Fun4Sim.C` as follows;
```
  SQChamberRealization* cham_real = new SQChamberRealization();
  cham_real->SetChamEff(0.5, 0.5, 0.5, 0.5, 0.5);
  se->registerSubsystem(cham_real);

...

reco->set_evt_reducer_opt("o");
```

### DIFF
--- a/framework/phool/PHFlag.cc
+++ b/framework/phool/PHFlag.cc
@@ -358,33 +358,33 @@ void PHFlag::WriteToFile(const string &name)
   map<string, int>::const_iterator intiter;
   for (intiter = intflag.begin(); intiter != intflag.end(); ++intiter)
   {
-    outFile << "I" << intiter->first << "\t" << intiter->second << endl;
+    outFile << "I\t" << intiter->first << "\t" << intiter->second << endl;
   }
 
   map<string, float>::const_iterator floatiter;
   for (floatiter = floatflag.begin(); floatiter != floatflag.end(); ++floatiter)
   {
-    outFile << "F" << floatiter->first << "\t" << floatiter->second << endl;
+    outFile << "F\t" << floatiter->first << "\t" << floatiter->second << endl;
   }
 
   int oldprecision = outFile.precision(15);
   map<string, double>::const_iterator doubleiter;
   for (doubleiter = doubleflag.begin(); doubleiter != doubleflag.end(); ++doubleiter)
   {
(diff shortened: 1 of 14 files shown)


abinashpun:
Dear Kenichi,

Thank you for the improvements and changes. For this PR too, I did sanity check by successfully building it and going over the code. I found a small bug in the GeomSvc code though it is not introduced by this PR. The L980 and L981 both have `planes[i].resolution = resol;` statement. 
https://github.com/E1039-Collaboration/e1039-core/blob/sim_consts/packages/geom_svc/GeomSvc.cxx#L981

Abi

knakano0524:
Thank you for the careful checks.  I have fixed the bug.  The array index in L981 is now `i + 1`.

> Also, I was wondering if we can introduce any fixed value of chamber resolution just like the efficiency value.
>
We can introduce a fixed value per chamber plane (which is what E906 used).  Or we can introduce a set of drift-distance-dependent values.  Indeed a temporary set is available as stored by `CalibParamXT`, but we need modify the tracker to use it in the reconstruction.  I think it is not necessary/urgent since the final track resolution is dominated by the FMag multiple scattering.

PR #106, opened 2021-07-03
More accurate/flexible requirement of geometric acceptance

knakano0524:
This PR is to add a new SubsysReco module, `SQGeomAcc`, which is a more accurate/flexible version of `RequireParticlesInAcc`.

`SQGeomAcc` has a mode for muon condition.  The mode can be `SINGLE`, `SINGLE_T` (muon at top), `PAIR`, `PAIR_TBBT` (pair at T+B or B+T), etc.  `SQGeomAcc` also has a mode for planes that define the acceptance.  The mode can be either `HODO`, `CHAM` or `HODO_CHAM` (both).

`SQGeomAcc` uses `SQHit` to judge whether a track hits a plane, where `RequireParticlesInAcc` uses `G4Hit`.  Therefore `SQGeomAcc` is more accurate about the acceptance boundary, although it is slower since the digitization has to be always carried out.  `RequireParticlesInAcc` was renamed to `SQGeomAccLoose`.  It might be still useful since it is faster but might be deleted in future.

I modified `e1039-analysis/SimChainDev/Fun4Sim.C` to make use of `SQGeomAcc`, which is commented out by default.  I used it and confirmed that `SQReco` reconstructed one or two muons in almost all accepted events when the `HODO_CHAM` mode and the `SINGLE` or `PAIR` mode were selected.

### DIFF
--- a/packages/UtilAna/UtilDimuon.cc
+++ b/packages/UtilAna/UtilDimuon.cc
@@ -21,6 +21,7 @@ SQDimuon* UtilDimuon::FindDimuonByID(const SQDimuonVector* vec, const int id_dim
   return 0;
 }
 
+/// OBSOLETE: Use `CalcVar()` instead.
 void UtilDimuon::GetX1X2(const SQDimuon* dim, double& x1, double& x2)
 {
   const double M_P    = 0.938;
@@ -33,6 +34,7 @@ void UtilDimuon::GetX1X2(const SQDimuon* dim, double& x1, double& x2)
   x2 = (p_beam  *p_sum)/(p_beam  *p_cms);
 }
 
+/// OBSOLETE: Use `CalcVar()` instead.
 void UtilDimuon::GetX1X2(const SQDimuon* dim, float& x1, float& x2)
 {
   double x1d, x2d;
@@ -41,6 +43,7 @@ void UtilDimuon::GetX1X2(const SQDimuon* dim, float& x1, float& x2)
   x2 = x2d;
 }
 
+/// OBSOLETE: Use `CalcVar()` instead.
 double UtilDimuon::GetX1(const SQDimuon* dim)
 {
   double x1, x2;
@@ -48,6 +51,7 @@ double UtilDimuon::GetX1(const SQDimuon* dim)
   return x1;
 }
 
+/// OBSOLETE: Use `CalcVar()` instead.
 double UtilDimuon::GetX2(const SQDimuon* dim)
 {
   double x1, x2;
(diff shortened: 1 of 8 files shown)


liuk:
Hi Kenichi, thanks for the fast implementation. I reviewed the new module and believe it should work as intended. I have one comment though, currently as long as there is one hit registered in a chamber, it would be considered in-acceptance. However at this time the reconstruction requires a track must have at least one hit per view to be considered valid, should we make the same requirement in the acceptance module?

knakano0524:
@liuk, Thank you for looking into the code.  You are correct that only D0X, D2X, D3pX and D3mX (not Xp, U(p) nor V(p)) are used in defining the chamber acceptance.  I think it is simple and sufficient in terms of the "geometric" acceptance, which this module considers.  D0U/V and D2U/V are identical to D0X and D2X in the plane size.  D3p/mU/V are larger than D3p/mX but the overlapping region of X, U and V is almost identical to X.

The tracking requirement that you mentioned can be regarded as a factor of the reconstruction efficiency, together with the chamber detection efficiency.  I would like to separate it from this module.  For example, we can use `SQChamberRealization` and `SQReco` to evaluate the effect of the efficiencies.

How do you think?  I will anyway add an explanation to the class documentation about the coverage of this module that I wrote above.

abinashpun:
Dear Kenichi,
Thank you for the changes. I looked at the code and built it successfully. Everything looks fine to me. Just to add on to what Kun mentioned for the tracking requirement, current reconstruction framework also ignores anything which passes through both top and bottom halves of drift chambers of station 3.

knakano0524:
@abinashpun; Thank you for your checks.  Do you mean T+T and B+B pairs are ignored?  I'm not sure if such selection is explicitly made (in VertexFit?) but anyway it is what we can control outside of this module.

I'd like to merge this PR tonight if you have no more comment.

liuk:
Hi Kenichi, I agree with your suggestion that these are more suitable for reconstruction efficiency rather than acceptance. I have no other suggestions and we should move forward with merging this PR.

liuk:
> Dear Kenichi,
> Thank you for the changes. I looked at the code and built it successfully. Everything looks fine to me. Just to add on to what Kun mentioned for the tracking requirement, current reconstruction framework also ignores anything which passes through both top and bottom halves of drift chambers of station 3.

I believe this means the station-3+ and 3- have a very small overlapping area that is currently not handled properly in tracking. This should be categorized as recon efficiency as well.

PR #107, opened 2021-07-13
Absolute cross sections of J/psi and psi' by E906 legacy generator and more

knakano0524:
The primary update in this PR is that the E906 legacy generator now properly sets the absolute scale of the J/psi and psi' cross sections.  Attached is a plot of the J/psi cross section, drawn by a new test macro (`CheckParticleGenJPsi.C`).
xsec_gen_jpsi

I changed the cross-section unit from nb to pb returned by `SQPrimaryParticleGen` since pb is used by PYTHIA8.

I made `SQPrimaryParticleGen` and others output more parameters to `recoConsts`.

I also modified `DbSvc` so that it now tries the DB connection 5 times at max, because we saw the connection sometimes fails in the online decoding and the grid job.

### DIFF
--- a/generators/E906LegacyGen/SQPrimaryParticleGen.C
+++ b/generators/E906LegacyGen/SQPrimaryParticleGen.C
@@ -12,6 +12,7 @@ from Kun to E1039 experiment in Fun4All framework
 #include <cstdlib>
 
 #include <fun4all/Fun4AllReturnCodes.h>
+#include <phool/recoConsts.h>
 #include <phool/PHCompositeNode.h>
 #include <phool/PHIODataNode.h>
 #include <phool/PHRandomSeed.h>
@@ -73,6 +74,7 @@ SQPrimaryParticleGen::SQPrimaryParticleGen(const std::string& name):
   _DrellYanGen(false),
   _JPsiGen(false),
   _PsipGen(false),
+  _vertexGen(0),
   ineve(NULL),
   _configFile("phpythia8.cfg"),
   _evt(0),
@@ -80,6 +82,8 @@ SQPrimaryParticleGen::SQPrimaryParticleGen(const std::string& name):
   _vec_dim(0),
   _integral_node(0),
   _dim_gen(new SQDimuon_v1()),
+  _pdfset("CT10nlo"),
+  _pdf(0),
   _pT0DY    (2.8),
   _pTpowDY  (1./(6. - 1.)),
   _pT0JPsi  (3.0),
@@ -89,23 +93,23 @@ SQPrimaryParticleGen::SQPrimaryParticleGen(const std::string& name):
   _weight_sum(0),
   _inte_lumi(0)
 {
- 
-  _vertexGen = new SQPrimaryVertexGen();
-  pdf = LHAPDF::mkPDF("CT10nlo", 0);
-
+  ;
 }
 
 
 SQPrimaryParticleGen::~SQPrimaryParticleGen()
 {
-  delete _dim_gen;
-  delete _vertexGen;
-  //delete pdf;
+  if (_dim_gen  ) delete _dim_gen;
+  if (_vertexGen) delete _vertexGen;
+  //delete _pdf;
 
 }
 
 int SQPrimaryParticleGen::Init(PHCompositeNode* topNode)
 {
+  _vertexGen = new SQPrimaryVertexGen();
+  _pdf = LHAPDF::mkPDF(_pdfset, 0);
+
   if(_PythiaGen){
   if (!_configFile.empty()) read_config();
   unsigned int seed = PHRandomSeed();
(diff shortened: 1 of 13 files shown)


generators/E906LegacyGen/SQPrimaryParticleGen.C
```diff
@@ -437,13 +453,14 @@ int SQPrimaryParticleGen::read_config(const char *cfg_file)
...
+ * The formula of pTmaxSq is discussed in DocDB 9292.
  */
 bool SQPrimaryParticleGen::generateDimuon(double mass, double xF)
 {
     double pz = xF*(DPGEN::sqrts - mass*mass/DPGEN::sqrts)/2.;
     
-    double pTmaxSq = (DPGEN::s*DPGEN::s*(1. - xF*xF) - 2.*DPGEN::s*mass*mass + mass*mass*mass*mass)/DPGEN::s/4.;
-  
+    //double pTmaxSq = (DPGEN::s*DPGEN::s*(1. - xF*xF) - 2.*DPGEN::s*mass*mass + mass*mass*mass*mass)/DPGEN::s/4.;
+    double pTmaxSq = (DPGEN::s / 4) * (1 - mass*mass/DPGEN::s)*(1 - mass*mass/DPGEN::s) * (1 - xF*xF);
```
knakano0524, line 463: I forgot to mention one important change.  Here I have changed the definition of `pTmaxSq` to follow the discussion made in E906.  Actually the previous definition was problematic in the E1039 code, since it sometimes became negative.

knakano0524:
I modified `SQPrimaryVertexGen` to accept two more modes (in addtion to "TargetOnly" and "DumpOnly"), as we discussed via e-mail.  "TargetDumpGap" enables only the air gap between target and dump.  "Manual" enables a z-range selected via `recoConsts` (`VTX_GEN_Z_START` and `VTX_GEN_Z_STOP`).

I stopped using the hard-coded `z_start` and `z_stop` values for the target and dump only modes.  Instead `geoManager` is used to find the z range for target, dump or gap in a new function, `FindMaterialRange()`.

I would propose to delete `set_targetOnlyMode()` and `set_dumpOnlyMode()` and use only `recoConsts` to control the mode, because the existence of the two control methods will make this class complicated.  Are you fine with it?

knakano0524:
When I set the mode to "Target" via the following line
```
  rc->set_CharFlag("VTX_GEN_MATERIAL_MODE", "Target");
```
The z range becomes -303.950...-296.050 cm.  In similar ways, the z range becomes 0.000...502.920 cm for "Dump" and -296.050...25.400 cm for "TargetDumpGap".

Any z range can be chosen in macro, by the following lines for example.
```
  rc->set_CharFlag("VTX_GEN_MATERIAL_MODE", "Manual");
  rc->set_DoubleFlag("VTX_GEN_Z_START", -350);
  rc->set_DoubleFlag("VTX_GEN_Z_STOP" ,   50);
```

liuk:
Thanks for the new interfaces for the vertex generator. I reviewed the changes and new interfaces and found no problems. I will leave the E906LegacyGen part for Abi to comment on.

knakano0524:
@liuk Thank you for the prompt review.  So I understand you are fine with removing the two interface functions (`set_targetOnlyMode()` and `set_dumpOnlyMode()`).  I made such change just now.  Also I added a Doxygen-style description.

generators/E906LegacyGen/SQPrimaryParticleGen.C
abinashpun: Thank you for the change. I also encountered the unphysical pTSq values in the past.

abinashpun:
Dear Kenichi,

Thank you for all the changes. I looked at the changes and also did sanity check by building it successfully. Also, I have committed a bug fix in the beam profile function of vertex generator code.

I was wondering if we have to add all those new recoConsts flags you added in this PR to the recoConsts.C file ( don't we have to add those if we want to use that from running macro ?).

knakano0524:
@abinashpun Thank you for looking into details.  That is a right question related to our strategy about the recoConsts flags.  It's not necessary to pre-define every flag in `recoConsts.C`.  Any flag can be newly defined in macro.  A difficulty is that the getters (like `get_CharFlag()`) causes a fatal error when the requested flag doesn't exist.  Therefore we better pre-define a set of key flags in `recoConsts.C`.

Regarding the new flags, `SQPrimaryVertexGen::init()` first checks if the "mode" flag is defined or not;
```
if (rc->FlagExist("VTX_GEN_MATERIAL_MODE"))
```
Thus no error occurs even when the flag is not defined.  I would prefer this method here, in order not to make `recoConsts.C` too long.

abinashpun:
Dear Kenichi,
Thank you for the explanation. I don't have any further comments. Please go ahead and merge the PR.

generators/E906LegacyGen/SQPrimaryParticleGen.C
Hugo-Leung: For future reference. It is true that the old definition of `pTmaxSq` will return negative values. But this is not a problem if a consistent definition of  kinematic variable. Using a consistent definition, when the previous definition of  `pTmaxSq` is negative, xT and/or xB would be outside of [0,1].

PR #108, opened 2021-08-05
Removal of dependency on PYTHIA6

knakano0524:
Our software uses PYTHIA8 in almost all cases.  Only `PHG4Decayer` depends on PYTHIA6.  This decayer is enabled only when
`PHG4Reco::set_decayer_active()` or `PHG4Reco::set_force_decay()` is called.  And actually these functions are never called in `e1039-core` nor `e1039-analysis`. 

On the other hand, Abi found that PYTHIA6 in `e1039-share` cannot be build at present, because it is now impossible to download the helper script (`build_pythia6.sh.txt`) from the Fermilab site without SSO.  There are several ways to sidestep this issue, but I think it is a good opportunity to remove PYTHIA6 from our software for simplicity.  If someone needs the function of `PHG4Decayer` in a future, we should implement it using PYTHIA8 or revive `PHG4Decayer` and PYTHIA6.

 I have confirmed that the updated code can be built and run fine, using `e1039-analysis/SimChainDev`.


abinashpun:
Dear Kenichi,

Thank you so much for the change and test. I went through the changes and they looked fine to me.

PR #109, opened 2021-08-13
New formula for tracklet charge and change in decoding v1495 data

knakano0524:
This PR includes four small changes.

1. The formula used in `FastTracklet::getCharge()` was modified in order to handle the anti-parallel magnet polarity.  It has been studied in DocDB 9505 and is explained in the Doxygen-style comment on the function.
2. The Main DAQ decoder temporarily skips the data from V1495 boards, because the data format is being changed because of the upgrade for the buffered readout.  After the upgrade is completed, the decoder will be modified again for the new data format.
3. The event weight in `SQMCEvent` is set to 1 by default.
4. `DbSvc` prints out a message when it retries a DB connection, for debugging purpose.

I have confirmed that the updated code can be built fine and is running fine for the online decoding.  The new charge formula has been tested by Forhad as explained in DocDB 9505.

### DIFF
--- a/interface_main/SQMCEvent_v1.cxx
+++ b/interface_main/SQMCEvent_v1.cxx
@@ -5,7 +5,7 @@ using namespace std;
 SQMCEvent_v1::SQMCEvent_v1()
   : _proc_id(0)
   , _xsec(.0)
-  , _weight(.0)
+  , _weight(1.0)
 {
   for (int ii = 0; ii < 4; ii++) {
     _par_id [ii] = 0;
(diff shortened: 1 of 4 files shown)


liuk:
Hi Kenichi, I have looked through the changes. The actual changes are simple enough and we should move forward with the merging.

forhadnmsu:
The charge definition worked as the details mentioned in the [docdb:9505](https://seaquest-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=9505&filename=charge.pdf&version=1).

PR #110, opened 2021-08-20
Update on PHG4SimpleEventGenerator

knakano0524:
This PR introduces mainly two updates on `PHG4SimpleEventGenerator`.
1. It creates the SQEvent and SQMCEvent nodes.  Particularly the run/spill/event IDs are now meaningful, which were always max. integer.
1. It sets `recoConsts` variables (like `PHG4SEG_EVENT_COUNT`) to record the number of generated events and the event-generator parameters.

There are two more small updates.
1. `SIGX_BEAM` in `recoConsts.cc` was changed to a realistic value (2.0 -> 0.3 cm).  I'd like to make all the values in `recoConsts.cc` realistic and remove the calls of `set_DoubleFlag()` etc. from `Fun4Sim.C`, in order to simplify the macro.  Do you think this course of action fine?
1. A potential problem in `SQPrimaryParticleGen::InitRun()` was fixed.  All the data nodes are now found or created regardless of the result of `if (!ineve)`.

I have confirmed that the update code can be built fine and the output of `PHG4SimpleEventGenerator` is as explained above.

### DIFF
--- a/framework/phool/recoConsts.cc
+++ b/framework/phool/recoConsts.cc
@@ -88,8 +88,8 @@ void recoConsts::set_defaults()
 
   set_DoubleFlag("X_BEAM", 0.);
   set_DoubleFlag("Y_BEAM", 0.);
-  set_DoubleFlag("SIGX_BEAM", 2.);
-  set_DoubleFlag("SIGY_BEAM", 2.);
+  set_DoubleFlag("SIGX_BEAM", 0.3);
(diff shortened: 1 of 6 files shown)


abinashpun:
Dear Kenichi,

The changes looks fine to me. Unless if you want me to test anything specific, you can merge the PR from my side.

knakano0524:
@abinashpun , Thank you for checking the changes.  I'd like to merge this PR by myself soon, in an hour or two.

PR #113, opened 2021-09-01
Update PHG4CylinderDetector.cc

zulkaidaakbar:
I modified the packing fraction from 90% to 60%

### DIFF
--- a/simulation/g4detectors/PHG4CylinderDetector.cc
+++ b/simulation/g4detectors/PHG4CylinderDetector.cc
@@ -73,8 +73,8 @@ void PHG4CylinderDetector::Construct(G4LogicalVolume *logicWorld)
 		// M_He = 0.4*0.145 = 0.058 g; M_NH3 = 0.6*0.867 = 0.520 g, M/rho = 0.578
 		// f_M_He = 10.0%, f_M_NH3 = 90.0%
 		G4Material* Target = new G4Material(name = "Target", density = 0.578 * g/cm3, ncomponents = 2);
-		Target->AddMaterial(sNH3, 90 * perCent);
-		Target->AddMaterial(lHe,  10 * perCent);
+		Target->AddMaterial(sNH3, 60 * perCent);
+		Target->AddMaterial(lHe,  40 * perCent);
 
 		TrackerMaterial = Target;
 


zulkaidaakbar:
We change the packing fraction from 90% to 60% since 90% packing fraction is not realistic. 60% is the typical packing fraction for NH3-beads target immersed in liquid Helium. We did CAT scan at UVA using plastic bead that mimic NH3 bead and we get really close to 60%

liuk (approved):

liuk:
Sorry I clicked the approve button too fast.

If we think 60% packing fraction is still a number that might change in the future, maybe we should provide a interface to allow user to change it instead of hard-coding it?

knakano0524:
@zulkaidaakbar Can you update the numbers described in the comment block above?

By writing this message, I realized that the quantity expected here is the percentage in *weight* and thus the orignal numbers "90:10" should be correct.  I understand "60:40" are the percentage in *volume*.

zulkaidaakbar:
Hi All,

I am sorry for not checking. I just checked and @knakano0524 is correct. The percentage in Geant AddMaterial is mass fraction and not packing (volume) fraction. I calculated and found that the mass fraction for lHe is 10.03%. So, I guess the present number (10% and 90%) is correct. 

@liuk : I think we will do one CAT-scan measurement so there will be one time change in the future. Maybe hard-coded is still fine. 

@knakano0524 :  "Can you update the numbers described in the comment block above?" I am sorry, what number in what comment block?

knakano0524:
I meant the comment block just above the lines you changed;
```
// M_He = 0.4*0.145 = 0.058 g; M_NH3 = 0.6*0.867 = 0.520 g, M/rho = 0.578
// f_M_He = 10.0%, f_M_NH3 = 90.0%
```
You no longer need to update them.

I remember the UVA group wants to change the material to completely different objects, like Air and 100% LHe.  Such flexible interface might be still wanted.

zulkaidaakbar:
Sorry, I accidentally clicked the request to re-review. 
@knakano0524 , thanks for inviting me to be a member. But I still don't see a button to assign reviewer here.

liuk:
It seems we don't need to change the packing fraction for now. Should I just close this PR?

zulkaidaakbar:
Yes, I we can close this pull request. Thank you

PR #114, opened 2021-09-21
Spill-by-spill SRawEvent output for online reconstruction

knakano0524:
This update is to introduce a spill-by-spill output of the decoded data for the online reconstruction.  The updated code is already running fine for the online decoding.

`Fun4AllSRawEventOutputManager` is the front end of the new output scheme.  It is enabled by default in the online decoding, i.e. `online/macros/Fun4MainDaq.C`.

The translation from SQEvent/Hit to SRawEvent will be useful for other classes.  Therefore I'd like to export the contents of `SQReco::BuildSRawEvent()` to a new namespace, `UtilSRawEvent`.  One problem here is that two member variables of `SQReco`, i.e. `_m_hitID_idx` and `_m_trghitID_idx`, cannot be handled by `UtilSRawEvent`.  But these variables seem not in use at present.  Can I remove them and use `UtilSRawEvent` in `BuildSRawEvent()`?

### DIFF
--- a/online/macros/Fun4MainDaq.C
+++ b/online/macros/Fun4MainDaq.C
@@ -4,6 +4,7 @@ R__LOAD_LIBRARY(libinterface_main)
 R__LOAD_LIBRARY(libdecoder_maindaq)
 R__LOAD_LIBRARY(libonlmonserver)
 R__LOAD_LIBRARY(libpheve_modules)
+R__LOAD_LIBRARY(libktracker)
 #endif
 
 int Fun4MainDaq(const int run=46, const int nevent=0, const bool is_online=false)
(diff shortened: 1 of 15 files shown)


liuk:
Hello Kenichi, I went through the code change and things make very good sense to me. 

Regarding removing SQReco::BuildSRawEvent(), the two hit index maps `_m_hitID_idx` and `_m_trghitID_idx` are actually needed if we would like have have `EventReducer` propagate the changes of hit information (like intime flags, and/or drift distance) back to the `SQHit`s. Maybe we could have `UtilSRawEvent::SetHit` and `UtilSRawEvent::SetTriggerHit` return the hit id maps?

knakano0524:
@liuk Thank you for looking into the changes.  I made a new commit to make `UtilSRawEvent` handle the hit ID maps, as you suggested.

liuk:
Thanks for the changes. Now everything looks good.  Please go ahead with the merge.

In reviewing the code, I noticed a bug introduced by a previous commit. I took the liberty of fixing that bug as well.

liuk (approved):

PR #115, opened 2021-09-21
Switches for filtering and digitization of simulated data

knakano0524:
Here is a small update to add switches for handling of simulated data.  I needed them to efficiently generate open-charm events.

`HepMCNodeReader` is already able to filter out particles by particle ID.  I added a function, `enable_position_filter(), to select particles by production position.  Using this function, we can have an event generator (like PYTHIA) produce muons and then `HepMCNodeReader` exclude muons produced behind FMag.

`SQDigitizer` doesn't digitize secondary particles by default.  I added a function, `set_digitize_secondaries()`, to change the behavior.

They won't change the default behavior of our simulation.  Thus this PR can be merged safely.

### DIFF
--- a/simulation/README.md
+++ b/simulation/README.md
@@ -1,0 +1,22 @@
+# Notes on E1039 Simulation Codes
+
+The E1039 simulation codes had been taken from (s)PHENIX, together with the Fun4All framework.
+Many obsolete/unused files are left in this directory.
+This note is to pick up and explain key files/classes, as a guideline for new developer.
+
+## Detector Configuration
+
+* The E1039 detector planes are built as `PHG4BlockSubsystem`, as defined in `macros/G4_SensitiveDetectors.C`.
+
+## Data Nodes
+
+Each event generator fills generated events into the "PHG4*" objects.
+Or some generators (like PYTHIA8) fills into the "HepMC" objects and then `HepMCNodeReader` transfers into "PHG4*".
+
+* `PHG4InEvent`
+    * `PHG4Particle`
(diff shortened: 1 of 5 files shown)


liuk:
Hello Kenichi, I went through the code changes and find everything reasonable. The only point I would like to discuss is the changes to the `SQDigitizer`, I'm wondering whether we should make this option `true` be default. What do you think?

abinashpun:
Dear Kenichi,

Thank you for the changes. When I was working with secondary hits, I found some very low momentum hits (~0 GeV) which gives the NAN values to slopes in digitization process. This results to the peak in the element ID =1 in hodoscope hit distribution ([DOCDB 9573](https://seaquest-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=9573&filename=bkg_23Sep21.pdf&version=1)).   If we consider making the the digitization of secondary hits true by default then I think we need to add some condition to bypass such hits.

knakano0524:
@liuk and @abinashpun Thank you for the comments.  I agree that we should consider secondary particles, particularly when studying chamber hits (where delta ray creates hit cluster).  But it seems safer to keep this option `false` by default at present.

I expect Abi will study the effect of secondary particles in his full background simulator.  We can discuss the default setting when we merge his updates.

liuk:
I agree. I don't have further comments. Let's keep this in mind and merge this PR for now.

liuk (approved):

abinashpun (approved):

PR #118, opened 2021-11-15
Inclusive Background simulation for E1039

abinashpun:
This pull request contains following major changes for implementation of the background study as envisioned by Kun and David. 

1. Implementation of the Legacy vertex generators in the PHPythia8 generator. Files changed for this purpose are; `generators/PHPythia8/PHPythia8.C`,`generators/PHPythia8/PHPythia8.h`, ` generators/phhepmc/PHHepMCGenHelper.cc` and `generators/phhepmc/PHHepMCGenHelper.h`
2. Adding filters in the `HepMCNodeReader` for selecting the background candidates. 
3. Modification and Implementation of `SQExternalGen` as `SQPileupGen` for pileup purpose. Three different methods of pileup were provided; fixed number, function, histogram.

More details in the technical report (hopefully within couple of days).

Also, the `SQPrimaryVertexGen` changes should be identical with the default one we have in the software. But this is giving us conflict. I will try to resolve that.


generators/PHPythia8/PHPythia8.C
```diff
@@ -41,17 +40,21 @@ PHPythia8::PHPythia8(const std::string &name)
...
-    return;
-  }
+    {
+      cout << "PHPythia8::Could not find $PYTHIA8 path!" << endl;
+      return;
+    }
 
   std::string thePath(charPath);
   thePath += "/xmldoc/";
   _pythia = new Pythia8::Pythia(thePath.c_str());
```
knakano0524, line 54: Is `new` here necessary?  It seems to me that `_pythia` will point to `ppGen` or `pnGen` later.

generators/PHPythia8/PHPythia8.C
```diff
@@ -86,22 +89,36 @@ int PHPythia8::Init(PHCompositeNode *topNode)
...
+    }
+
+  //if(! _legacy_vertexgenerator) _pythia->init();
+  //else
   {
-    cout << PHWHERE << " ERROR: seed " << seed << " is not valid" << endl;
-    exit(1);
-  }
+    ppGen->readString("Beams:idB = 2212");
+    pnGen->readString("Beams:idB = 2112");
```
knakano0524, line 117: It seems to always overwrite the settings read from the config file.  If so, when `_legacy_vertexgenerator` is false, does PHPythia8 will work properly as it does?

generators/PHPythia8/PHPythia8.C
```diff
@@ -249,12 +279,13 @@ int PHPythia8::process_event(PHCompositeNode *topNode)
...
-    _integral_node->set_N_Generator_Accepted_Event(_pythia->info.nAccepted());
-    _integral_node->set_N_Processed_Event(_eventcount);
-    _integral_node->set_Sum_Of_Weight(_pythia->info.weightSum());
-    _integral_node->set_Integrated_Lumi(_pythia->info.nAccepted() / (_pythia->info.sigmaGen() * 1e9));
-  }
+    {
+      _integral_node->set_N_Generator_Accepted_Event(_pythia->info.nAccepted());
+      _integral_node->set_N_Processed_Event(_eventcount);
+      _integral_node->set_Sum_Of_Weight(_pythia->info.weightSum());
+      _integral_node->set_Integrated_Lumi(_pythia->info.nAccepted() / (_pythia->info.sigmaGen() * 1e9));
```
knakano0524, line 286: Are these variables correct in case of both `_legacy_vertexgenerator` = true and false?

knakano0524 (changes requested):
@abinashpun Thank you for this large amount of updates.  The background simulation is definitely important.

I have looked at only `PHPythia8` so far.  I'd like to make sure that it will work as it does now when the legacy vertex generator is off.  I made a few inline comments.

generators/PHPythia8/PHPythia8.C
abinashpun: In config file, we have been just defining only Beams:idA as  `Beams:idA= 2212` but not Beams:idB. If we don't define it then I suppose the Beam:idB becomes same as Beam:idA. But here, we are defining two different Beams:id to choose from pp vs pn.

generators/PHPythia8/PHPythia8.C
abinashpun: Yes, they work for both cases.

generators/PHPythia8/PHPythia8.C
knakano0524: I see.  I confirmed that `idB` (and also `idA`) are 2212 by default;
https://pythia.org/latest-manual/BeamParameters.html

generators/PHPythia8/PHPythia8.C
knakano0524: I wonder how they work.  I looked at only the diffs, but I understand `_pythia` points to `ppGen` or `pnGen`, and thus only one of them is  here.  But do we need both (i.e. sum) of them?

simulation/g4main/HepMCNodeReader.h
```diff
@@ -59,6 +59,11 @@ class HepMCNodeReader : public SubsysReco
 
   bool PassParticleFilter(HepMC::GenParticle * p);
 
+  ///@some added filter interface for bkg decay study
+  void Set_bkg_mode(){_bkg_mode = true;}// abi
```
knakano0524, line 63: Could you add a Doxygen comment to explain the meaning of this mode?  The contents are quite complicated, and the function name seems not well explanatory.  You could refer to your document on DocDB.

generators/E906LegacyGen/SQPileupGen.C
```diff
@@ -0,0 +1,195 @@
...
+{
+  if(_beam_intensity_profile) 
+    {
+      setBucketSize((int)_beam_intensity_profile->GetRandom()/_proton_coeff);//if the function describes dN/dI vs I   
+    }
+
+  if(_beam_intensity_profile_histo)
+    {
+      double QIEcount = _beam_intensity_profile_histo->GetRandom();
+      if (QIEcount>_inhibit_threshold) return Fun4AllReturnCodes::ABORTEVENT;
```
knakano0524, line 137: Is it possible to apply this cut in case of `_beam_intensity_profile`?  It should be helpful when we compare profiles.

Is there a good reason of using `return` here, rather than repeating `GetRandom()` until `QIEcount < _inhibit_threshold`?

generators/E906LegacyGen/SQPileupGen.C
```diff
@@ -0,0 +1,195 @@
...
+  _n_proc_evt++;
+  _evt->set_run_id(0);
+  _evt->set_spill_id(0);
+  _evt->set_event_id(_n_proc_evt);
+  return Fun4AllReturnCodes::EVENT_OK; 
+}
+
+bool SQPileupGen::readExtTree(int nEvents)
+{
+  std::cout<<"nEvents: "<<nEvents<<std::endl;
```
knakano0524, line 181: Please add `if (Verbosity() > 0)` for example.

knakano0524 (commented):
I went through the other files.  The changes look fine overall, but I haven't looked into all details.  If you could tell me what I should check specifically, I would do that.

generators/PHPythia8/PHPythia8.C
abinashpun: Dear Kenichi. yes, `_pythia` point to either `ppGen` or `pnGen`. As these values are incremented event-wise, they both should have contribution towards the total events' integrated luminosity. Does that make sense?

simulation/g4main/HepMCNodeReader.h
abinashpun: I will do that. Thank you.

generators/E906LegacyGen/SQPileupGen.C
abinashpun: Thank you. I will follow the suggestion.

generators/E906LegacyGen/SQPileupGen.C
abinashpun: I was wondering if you mean to apply in the histogram itself.  Could you please make both of the comments more clear. Thank you.

generators/PHPythia8/PHPythia8.C
knakano0524: > they both should have contribution towards the total events' integrated luminosity.
>
I understand so.  Then the present code is not sufficient.  Instead the sum like `ppGen->info.nAccepted() + pnGen->info.nAccepted()` should be used.

generators/E906LegacyGen/SQPileupGen.C
knakano0524: I mean to apply the cut in case of TF1, when  `_beam_intensity_profile` is true.

The 2nd comment means the following change for example.  It could be improper in terms of normalization.
```
double QIEcount;
do {
  QIEcount = _beam_intensity_profile_histo->GetRandom();
} while (QIEcount>_inhibit_threshold);
```

abinashpun:
> I went through the other files. The changes look fine overall, but I haven't looked into all details. If you could tell me what I should check specifically, I would do that.

Dear Kenichi,
Thank you so much for your helpful comments. I was wondering if you could have a look into the conflict in the `SQPrimaryVertexGen` and the `HepMCNodeReader`

generators/E906LegacyGen/SQPileupGen.C
abinashpun: 1st comment: I see. It should be possible. `void TF1::SetRange(Double_t  xmin, Double_t  xmax)` with xmax being our threshold cut. 


2nd comment: It would be a loop within an event. I was wondering why we would want that loop within an event. I believe, we want to have a distribution of the QIE count over the run not over the event. Don't we?

generators/PHPythia8/PHPythia8.C
abinashpun: Dear Kenichi,
Thank you for the detail look. yes, it does seems like we need to set the integral node separately in case of `_legacy_vertexgenerator = true` by adding both from `ppGen` and `pnGen`

generators/PHPythia8/PHPythia8.C
abinashpun: Yes, I have deleted L54 as our discussion.

abinashpun:
Dear @knakano0524 ,

I have resolved the conflicts and addressed your comments. Please let me know if you have further comments.

Dear @liuk ,
Please let me know if you have any comments regarding the pull request.

generators/E906LegacyGen/SQPileupGen.C
```diff
@@ -0,0 +1,209 @@
...
+
+  if(_beam_intensity_profile) 
+    {
+      //setBucketSize((int)_beam_intensity_profile->GetRandom()/_proton_coeff);//if the function describes dN/dI vs I   
+      do {
+       QIEcount = _beam_intensity_profile->GetRandom();
+      } while (QIEcount>_inhibit_threshold);
+      
+      int  bucket_size = (int)QIEcount/_proton_coeff;
+      setBucketSize(bucket_size);
```
knakano0524, line 139: Move them to line 154 and delete lines 151-152?

generators/E906LegacyGen/SQPileupGen.C
```diff
@@ -0,0 +1,209 @@
...
+      
+      int  bucket_size = (int)QIEcount/_proton_coeff;
+      setBucketSize(bucket_size);      
+    }
+
+  if(_beam_intensity_profile_histo)
+    {
+      //double QIEcount = _beam_intensity_profile_histo->GetRandom();
+      //if (QIEcount>_inhibit_threshold) return Fun4AllReturnCodes::ABORTEVENT;
+
```
knakano0524, line 146: Delete these?

generators/PHPythia8/PHPythia8.C
```diff
@@ -87,22 +90,36 @@ int PHPythia8::Init(PHCompositeNode *topNode)
...
+    }
   else
+    {
+      cout << PHWHERE << " ERROR: seed " << seed << " is not valid" << endl;
+      exit(1);
+    }
+
+  //if(! _legacy_vertexgenerator) _pythia->init();
+  //else
   {
```
knakano0524, line 116: Delete these?

generators/PHPythia8/PHPythia8.C
```diff
@@ -114,29 +131,29 @@ int PHPythia8::End(PHCompositeNode *topNode)
...
+      _pythia->stat();
+
+      //match pythia printout
+      cout << " |                                                                "
+	   << "                                                 | " << endl;
+      cout << "                         PHPythia8::End - " << _eventcount
+	   << " events passed trigger" << endl;
+      cout << "                         Fraction passed: " << _eventcount
+	   << "/" << _pythia->info.nAccepted()
+	   << " = " << _eventcount / float(_pythia->info.nAccepted()) << endl;
```
knakano0524, line 145: Can you correct these lines, using ppGen and pnGen?

generators/PHPythia8/PHPythia8.C
```diff
@@ -254,12 +284,26 @@ int PHPythia8::process_event(PHCompositeNode *topNode)
...
-    _integral_node->set_Integrated_Lumi(_pythia->info.nAccepted() / (_pythia->info.sigmaGen() * 1e9));
-  }
+    {
+      if(_legacy_vertexgenerator)
+            
+	{
+	  _integral_node->set_N_Generator_Accepted_Event(ppGen->info.nAccepted()+pnGen->info.nAccepted());
+	  _integral_node->set_N_Processed_Event(_eventcount);
+	  _integral_node->set_Sum_Of_Weight(ppGen->info.weightSum()+pnGen->info.weightSum());
+	  _integral_node->set_Integrated_Lumi((ppGen->info.nAccepted()+pnGen->info.nAccepted()) / ((ppGen->info.sigmaGen()+pnGen->info.sigmaGen()) * 1e9));
```
knakano0524, line 294: The right formula should be `ppGen->info.nAccepted() / ppGen->info.sigmaGen() + pnGen->info.nAccepted() / pnGen->info.sigmaGen()`.

generators/PHPythia8/PHPythia8.C
```diff
@@ -71,9 +74,9 @@ int PHPythia8::Init(PHCompositeNode *topNode)
 {
   if (!_configFile.empty()) read_config();
   for (unsigned int j = 0; j < _commands.size(); j++)
-  {
-    _pythia->readString(_commands[j]);
-  }
+    {
+      _pythia->readString(_commands[j]);
```
knakano0524, line 78: This line should fail since `_pythia = 0` here.  Probably we have to use both `ppGen` and `pnGen`.

generators/PHPythia8/PHPythia8.C
```diff
@@ -254,12 +284,26 @@ int PHPythia8::process_event(PHCompositeNode *topNode)
...
   if (_integral_node)
-  {
-    _integral_node->set_N_Generator_Accepted_Event(_pythia->info.nAccepted());
-    _integral_node->set_N_Processed_Event(_eventcount);
-    _integral_node->set_Sum_Of_Weight(_pythia->info.weightSum());
-    _integral_node->set_Integrated_Lumi(_pythia->info.nAccepted() / (_pythia->info.sigmaGen() * 1e9));
-  }
+    {
+      if(_legacy_vertexgenerator)
+
```
knakano0524, line 289: Delete this blank line?

simulation/g4main/HepMCNodeReader.cc
```diff
@@ -144,178 +148,284 @@ int HepMCNodeReader::process_event(PHCompositeNode *topNode)
...
+	      if (ishape == ShapeG4Tubs)
+		{
+		  if (sqrt(xpos * xpos + ypos * ypos) > worldsizey / 2 ||
+		      fabs(zpos) > worldsizez / 2)
+		    {
+		      if (Verbosity()>0)
+		      cout << "vertex x/y/z" << xpos << "/" << ypos << "/" << zpos
+			   << " outside world volume radius/z (+-) " << worldsizex / 2
+			   << "/" << worldsizez / 2 << ", dropping it and its particles"
+			   << endl;
```
knakano0524, line 274: Please make the line indent right.  I'd suggest to also use `{}` for multiple lines (even in case of a single command).

simulation/g4main/HepMCNodeReader.cc
```diff
@@ -144,178 +148,284 @@ int HepMCNodeReader::process_event(PHCompositeNode *topNode)
...
+	      else if (ishape == ShapeG4Box)
+		{
+		  if (fabs(xpos) > worldsizex / 2 || fabs(ypos) > worldsizey / 2 ||
+		      fabs(zpos) > worldsizez / 2)
+		    {
+		      if (Verbosity()>0)
+		      cout << "Vertex x/y/z " << xpos << "/" << ypos << "/" << zpos
+			   << " outside world volume x/y/z (+-) " << worldsizex / 2 << "/"
+			   << worldsizey / 2 << "/" << worldsizez / 2
+			   << ", dropping it and its particles" << endl;
```
knakano0524, line 287: The line indent here also.

knakano0524 (changes requested):
Here are my comments on the updated code.  Most of them are not critical, but we should keep the code of event generators clean because its logic will be sought by analyzers in future.

generators/E906LegacyGen/SQPileupGen.C
abinashpun: Thank you for the comment kenichi.

knakano0524 (changes requested):
Could you resurrect a few lines that you removed?  Or I want to know why they had to be removed.

generators/E906LegacyGen/SQPileupGen.C
```diff
@@ -0,0 +1,200 @@
...
+int SQPileupGen::process_event(PHCompositeNode* topNode)
+{
+  double QIEcount;
+  
+  if(_beam_intensity_profile) 
+    {
+      do {
+       QIEcount = _beam_intensity_profile->GetRandom();
+      } while (QIEcount>_inhibit_threshold);
+      _bucketSize  = (int)QIEcount/_proton_coeff;
```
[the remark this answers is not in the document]
abinashpun: Dear Kenichi,
Yes I thought you wanted me to delete the lines, because we were repeating same setBucketSize for the function and histogram case. I realized that we can simply set the _bucketSize value to take care of both case. Also, if we put that outside the loop then we will mess up while the case when set bucketsize by a constant number.

generators/PHPythia8/PHPythia8.C
```diff
@@ -111,32 +127,30 @@ int PHPythia8::End(PHCompositeNode *topNode)
 {
   if (verbosity >= VERBOSITY_MORE) cout << "PHPythia8::End - I'm here!" << endl;
 
-  recoConsts::instance()->set_IntFlag("PYTHIA8_EVENT_COUNT", _eventcount);
-
```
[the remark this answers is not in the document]
abinashpun: Thank you for detail look. Those were accidental deletes.

generators/PHPythia8/PHPythia8.C
```diff
@@ -151,14 +165,15 @@ int PHPythia8::read_config(const char *cfg_file)
...
-    exit(2);
-  }
-
-  _pythia->readFile(_configFile.c_str());
+    {
+      cout << "PHPythia8::read_config - Failed to open file " << _configFile << endl;
+      exit(2);
+    }
 
-  recoConsts::instance()->set_CharFlag("PYTHIA8_CONFIG_FILE", _configFile);
```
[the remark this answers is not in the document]
abinashpun: Thank you for detail look. Those were accidental deletes.

generators/E906LegacyGen/SQPileupGen.C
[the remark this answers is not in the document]
knakano0524: I see.  I first wanted to put the lines outside the loop, and I didn't take into account the constant number case.  You are correct and the present code is fine.  It is helpful if you could explain your thought when making such change in the future.

generators/PHPythia8/PHPythia8.C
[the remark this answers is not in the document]
knakano0524: The code is fine now.  I wonder how such accidental deletion happened and how you will prevent it.

knakano0524 (approved):

abinashpun:
Thank you Kenichi for your careful look and inputs. I will keep in mind about your comments and suggestion in future. 
@liuk , let me know if you have any comments for this PR.

PR #120, opened 2021-12-10
changed PHG4INEVENT from PHDataNode to PHIODataNode

abinashpun:
Changed PHINEVENT from PHDataNode to PHIODataNode to make it more flexible to use in the DST level analysis (as found by discussion with Kenichi). Most of the changes are very small in the generator levels. I built the change successfully.


knakano0524 (approved):
Thank you for making the pull request.  You could merge it now, as the change looks fine and you have already tested it.

PR #122, opened 2021-12-22
Small updates on SQPileupGen, MainDaqParser and CalibDriftDist

knakano0524:
I'd like to suggest a small update on SQPileupGen, together with other unrelated ones that I need now.

I added `_evt->set_qie_rf_intensity(0, QIEcount)` to `SQPileupGen.C`, which sets the QIE count to SQEvent.  It should be useful in comparing the piled-up events with the E906 NIM3 events etc.  I also changed `TH1D*` to `TH1*` in SQPileupGen, where we had better use the base class in general.

I modified the decoder (`MainDaqParser`), which now extracts the run-end time from the Coda End event.  The run-end time had been available but just not utilized by the decoder. 

I modified `CalibDriftDist` (and `RunParamBase`), which is now able to read a parameter set (i.e. chamber X-T curve) from a local file.  This function will be soon used by the calibration of the X-T curve using cosmic rays.

I added a "-B" option to `build.sh`.  It sets the build directory to `../core-build`, instead of `./build`.  I think it convenient because the source directory is kept clean.  Can I even make it default?

The updated code can be built fine.  I have only confirmed that the run-end time extracted by `MainDaqParser` is reasonable.  The other changes should not affect the default software behavior.  I will do individual tests in the coming weeks.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -21,7 +21,7 @@ install=$MY_INSTALL
 mode=all
 OPTIND=1
 cmake_args=""
-while getopts ":s:r:i:c:b:" OPT ; do
+while getopts ":s:r:i:c:b:B" OPT ; do
     case $OPT in
         s ) mode='single'
             package=$OPTARG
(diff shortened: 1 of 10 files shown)


abinashpun:
Dear Kenichi,

Thank you for the changes. I had a look to the changes and they seem fine to me. Also, I agree with making the source directory clean by moving the build directory outside of it.

knakano0524:
@abinashpun Thank you for the checks.  I think you need not do any specific test.  I'd like to merge this PR in one hour.  I will soon make another pull request, in which I will change the default build directory.

abinashpun (approved):

PR #123, opened 2021-12-23
Addition of OnlMonParam etc.

knakano0524:
This PR is to add a new utility class, `OnlMonParam`, which handles run-dependent parameters for the OnlMon client.  It is used by `OnlMonMainDaq` for now, which reads the expected number (=83) of TDC boards read out per event, and makes a warning if the number is found different in runs/events.  Parameters are stored in TSV files under `/data2/e1039/onlmon/param/`, which can be modified by subsystem experts at anytime.

I also imported `E906SpillSelector` and `FilterByTrigger` from `e1039-analysis/GenRoadset/src/`.  They should be useful for other online/offline analyses when selecting events by spill IDs (of E906 data for now) and trigger bits.  We can add such general-purpose SubsysReco modules to `evt_filter/` in the future.

The updated code can be compiled fine.  I have confirmed that it runs fine on online (i.e. real) data.  It should not affect the offline analysis.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -38,7 +38,7 @@ while getopts ":s:r:i:c:b:" OPT ; do
         c ) cmake_args=$OPTARG
             echo " - pass additional args $cmake_args to cmake"
             ;;
-        b ) build=$OPTARG
+        b ) build=$(readlink -e $OPTARG)
             echo "Build directory = $build"
             ;;
         * ) echo 'Unsupported option.  Abort.'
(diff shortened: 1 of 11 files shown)


knakano0524:
I forgot to mention that `build.sh` was changed slightly, but the default build directory is still `./build`.  Let me postpone the change that we discussed in PR 122, since it will cause a conflict in this PR.

I think you need not build and test the updated code.

abinashpun (approved):

PR #124, opened 2022-01-03
Arrangement of `CalibDriftDist`, `EventReducer` and `PHG4Reco`

knakano0524:
This PR includes two changes.

1. I modified `CalibDriftDist` (and `Fun4AllEVIOInputManager`).  It now reads the calibration parameter stored at `$E1039_RESOURCE/calib/xt_curve/20211222` (actually it reads a copy on MySQL DB) to update the drift distance, the in-time window and the element position of `SQHit`.  I dropped the 'e' option of `EventReducer` in order to make `CalibDriftDist` unique.  I plan to factor out other functions of `EventReducer`.

2. I added a new function, `PHG4Reco::set_field_map()`, which takes no argument.  It instead reads the parameters from `recoConsts`.  On the other hand, the existing function `set_field_map(const std::string &fmap, const PHFieldConfig::FieldConfigTypes dim)` requires the parameters to be written as string in each Fun4All macro, like `e1039-analysis/SimChainDev/Fun4Sim.C`.  Thus the new function makes Fun4All macros simpler.

I have confirmed that the update code can be built fine.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -15,7 +15,7 @@ test -z "$OFFLINE_MAIN" && echo "Need set 'OFFLINE_MAIN'.  Abort." && exit
 test -z "$MY_INSTALL"   && echo   "Need set 'MY_INSTALL'.  Abort." && exit
 
 src=$(dirname $(readlink -f $0))
-build=`pwd`/build
+build=$src/../core-build
 install=$MY_INSTALL
 
 mode=all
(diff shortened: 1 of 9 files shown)


knakano0524:
I committed an irrelevant but urgent change.  It is to fix the event weight of the D-Y, J/psi and psi' processes in the E906 legacy generator.  I had omitted the range of xF (and mass for D-Y) from the event weight somehow, which is clearly necessary and indeed included in the event weight of the E906 generator.

This change makes the ratio of J/psi to D-Y reasonable, where the ratio was found by Forhad too large at present.

abinashpun:
Daer Kenichi,

Thank you for the changes. I went over the changes and they seem fine to me.

abinashpun (approved):

PR #125, opened 2022-02-01
Updates on Main DAQ decoder, chamber X-T curve and chamber realization

knakano0524:
This update is primarily to enable the V1495 TDC readout of the Main DAQ decoder.  The present cosmic runs are being decoded by this updated version, and the online monitor is showing reasonable hit distributions.

This update also includes a change in the meaning of the chamber X-T curve parameter, which is handled by `CalibParamXT`.  Now its time (T) means the *TDC* time instead of the *drift* time, and the T range of the X-T curve is regarded as the in-time window.  `CalibParamInTimeTaiwan` now takes care of only the hodoscope parameters.  This new parameter handling should be easier to be maintained.  The parameter values under `e1039-resource` have been modified accordingly.

`SQChamberRealization` and `CalibDriftDist` were modified to use the updated `CalibParamXT` properly.  `SQChamberRealization` was modified also to resolve a problem found by DarkQuest, namely it now erases inefficient hits (instead of setting the in-time flag to `false`).  The previous method didn't work because the in-time flag was overwritten by `CalibDriftDist`.

I have confirmed that the updated version can be built for online and offline.

### DIFF
--- a/interface_main/SQHitVector.h
+++ b/interface_main/SQHitVector.h
@@ -54,6 +54,7 @@ public:
   virtual       SQHit* at(const size_t idkey) {return NULL;}
   virtual       void   push_back(const SQHit *hit) {}
   virtual       size_t erase(const size_t idkey) {return 0;}
+  virtual       Iter   erase(Iter pos) {return HitVector().end();}
 
   virtual ConstIter begin()                   const {return HitVector().end();}
   virtual ConstIter   end()                   const {return HitVector().end();}
--- a/interface_main/SQHitVector_v1.h
+++ b/interface_main/SQHitVector_v1.h
@@ -41,6 +41,7 @@ public:
 	  _vector.erase(_vector.begin() + idkey);
 	  return _vector.size();
 	}
+  Iter erase(Iter pos) { return _vector.erase(pos); }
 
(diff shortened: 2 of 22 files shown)


knakano0524:
I think you need not test the updated version because the changes are specific to the decoder and the chamber parameter.  Comments on the general concept are welcome.

I will create a package under `e1039-analysis` to test and demonstrate the use of the new chamber realization.

knakano0524:
I committed more changes about the chamber realization.

* `SQChamberRealization` now has more adjustable parameters, such as a fixed value and a scaling factor applied to the resolution.
* `CalibHodoInTime` and `CalibDriftDist` now take run ID from RUNNUMBER, instead of SQRun.  It should be better since `GeomSvc` is taking run ID from RUNNUMBER.
* `SQReco` now clears the contents of `TrackletVector` at the beginning of each event, as it is already doing for `SRecTrack` and `SRecDimuon`.  It is necessary when the Fun4All input (i.e. DST) contains a reconstructed result.
* The parameter structure in `CalibParamXT` was refined so that its usage gets simpler.

I would like to merge this PR at 10 PM today.  If you need more time for review, please let me know.

PR #128, opened 2022-03-09
New variables+canvases added to OnlMonMainDaq

knakano0524:
This update is mainly to add new variables and canvases to OnlMonMainDaq.  Attached are the new two canvases that show
 * The plane occupancy (i.e. N of hits/plane/event) and
 * The number of events, a decoding time and an analysis time per spill.
OnlMonMainDaq_can1
OnlMonMainDaq_can2
They should be useful in the DAQ stress test on Friday.  Moreover the attached images (which were made with run 3700) already show a interesting (strange) behavior around the 700th spill.  We should investigate this later.

The decoding time shown here includes the waiting time for new words in Coda file, which is actually dominant.  The real decoding time is <0.1 sec/spill, as seen in several spills.  The analysis time is defined as the time when the last event is processed by "process_event()" minus the time when the first event is processed.

To record these decoding and analysis times, I created two types of the DST data nodes; `SQHardSpill` and `SQHardEvent`.  They should contain hardware-related variables.  I moved several variables (such as Coda event ID, VME time, N of TDCs) from `SQEvent' to `SQHardEvent`.  In offline production, `SQHard*` need not be written out.

I have confirmed that the updated code can be built fine.  It is being used in the online decoding.

### DIFF
--- a/interface_main/SQEvent.h
+++ b/interface_main/SQEvent.h
@@ -9,7 +9,7 @@
 #define _H_SQEvent_H_
 
 #include <phool/PHObject.h>
-
+#include <climits>
 #include <map>
 #include <iostream>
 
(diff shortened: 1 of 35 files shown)


abinashpun (approved):

knakano0524:
@abinashpun Thank you for the check.

@liuk Actually I had better merge this PR by tomorrow noon, because Abi has to make use of the updated version for the semi-online reconstruction.  So could you take a look by noon?

liuk (approved):

PR #129, opened 2022-03-26
kTThreads merge

nwuerfel:
Merge of kTThreads package along with e1039 data input converter. Small edits to setup script to allow for environmental variables to be set properly on Tracking PC. Also changed TNode to SQTNode after conversation with Kun and previous compilation issues. Sample macro for running kTThreads included.


liuk:
The only change to the existing code seems to be a simple internal class name change., so I think we should be ready to merge.

However, shouldn't there be a change in the build.sh script to include the new package in the building process?

abinashpun:
Hi all,
If I understand correctly, KSchedular is a non-subsysco module and other e1039-core packages don't depend on it. I was wondering if it is better to put it in separate repository or may be in e1039-analysis area itself.

packages/kTThreads/KScheduler.cxx
```diff
@@ -0,0 +1,1097 @@
...
+ * _m_trghitID_idx
+ * _event_header
+ * _event
+ * _spill_map
+ * _triggerhit_vector
+ * _hit_vector
+ */
+
+// converter from e1039 SQHitVector to SRawEvent for ktracker
+SRawEvent* BuildSRawEvent(SQEvent* sqevent, SQHitVector* sqhitvector, SQHitVector* sqtrighitvector){
```
knakano0524, line 501: Can you make `BuildSRawEvent()` a member function of KScheduler?

packages/kTThreads/macros/old/kThreadTest.C
```diff
@@ -0,0 +1,50 @@
+
```
knakano0524, line 1: Do you want to keep this "old" macro/directory?  If so, can you rather rename it a proper name?

knakano0524 (commented):
The code looks good overall.  Here I just have a few minor comments.  As Kun mentioned, this package should be added to `build.sh`, and then please confirm that `build.sh` runs fine.

packages/kTThreads/macros/old/kThreadTest.C
nwuerfel: I will delete it, the removal commit was lost when I moved from a clone to fork directory.

nwuerfel:
> Hi all, If I understand correctly, KSchedular is a non-subsysco module and other e1039-core packages don't depend on it. I was wondering if it is better to put it in separate repository or may be in e1039-analysis area itself.

Hi Abinash, you're correct that KScheduler is non-subsysco module and nothing else in e1039-core depends on it. However, I don't think that it should be part of e1039-analysis because it will be specifically used for the online reconstruction effort. Having it included in e1039-core is a convenient way of distributing the package and many parts of KScheduler depend on other packages in e1039-core, so it is easier to build as well. If we need a separate repository, maybe we have an e1039-core and e1039-core-online but syncing them may be a pain.

abinashpun:
> > Hi all, If I understand correctly, KSchedular is a non-subsysco module and other e1039-core packages don't depend on it. I was wondering if it is better to put it in separate repository or may be in e1039-analysis area itself.
> 
> Hi Abinash, you're correct that KScheduler is non-subsysco module and nothing else in e1039-core depends on it. However, I don't think that it should be part of e1039-analysis because it will be specifically used for the online reconstruction effort. Having it included in e1039-core is a convenient way of distributing the package and many parts of KScheduler depend on other packages in e1039-core, so it is easier to build as well. If we need a separate repository, maybe we have an e1039-core and e1039-core-online but syncing them may be a pain.

I see. That sounds reasonable. I don't have any more comment. Please let me know if you want me check anything specific.

knakano0524:
@nwuerfel Thank you for the updates.  They look OK.  I think you can merge this PR now.

PR #130, opened 2022-04-19
Modified version of kTracker for local-tracklet reconstruction

knakano0524:
This update is to introduce a modified version of kTracker that reconstructs local tracklets using drift distance.  It is helpful in calibrating the X-T curve using cosmic data.  I'd like to first hear your opinions about the concept+method of this update.  If it looks fine, I will straighten up the code and ask you a closer review.

The original tracklet reconstruction (i.e. `KalmanFastTracking::buildTrackletsInStation()`) ignores the drift distance, as the drift sign of hits is not resolved at that stage.  Thus I created new classes; `SQTrackletReco` and `KalmanFastTrackletting`, by inheriting them from `SQReco` and `KalmanFastTracking`.  `KalmanFastTrackletting` takes all combinations of drift signs to reconstruct D3, D2 and D2+D3 tracklets.

`SQReco` and `KalmanFastTracking` were modified slightly so that the new classes can inherit them.  In addition, I replaced `_DEBUG_ON` with `verbosity`, and added a member function `SQReco::set_output_list_index()` (which is actually not helpful for my purpose but many helpful in future).

`GeomSvc` and `CalibDriftDist` were modified so that the plane resolution can be changed together with the X-T curve in `CalibDriftDist`.  I think `GeomSvc::getPlanePtr()` will be useful in getting/setting all plane parameters.

A change in `Fun4AllEVIOInputManager::run()` is just a bug fix.

### DIFF
--- a/online/decoder_maindaq/Fun4AllEVIOInputManager.cc
+++ b/online/decoder_maindaq/Fun4AllEVIOInputManager.cc
@@ -287,6 +287,7 @@ int Fun4AllEVIOInputManager::run(const int nevents)
 
   run_header->set_run_id         (rd->run_id);
   run_header->set_unix_time_begin(rd->utime_b);
+  run_header->set_unix_time_end  (rd->utime_e);
   for (int ii = 0; ii < 5; ii++) {
     run_header->set_fpga_enabled (ii, rd->fpga_enabled [ii]);
     run_header->set_fpga_prescale(ii, rd->fpga_prescale[ii]);
--- a/packages/calibrator/CalibDriftDist.cc
+++ b/packages/calibrator/CalibDriftDist.cc
@@ -19,6 +19,11 @@ CalibDriftDist::CalibDriftDist(const std::string& name)
   , m_fn_xt("")
   , m_vec_hit(0)
   , m_cal_xt (0)
+  , m_reso_d0 (0)
+  , m_reso_d1 (0)
+  , m_reso_d2 (0)
+  , m_reso_d3p(0)
+  , m_reso_d3m(0)
 {
   ;
 }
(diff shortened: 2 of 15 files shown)


abinashpun:
Dear Kenichi,
Thank you for the changes. Actually I still need to look more carefully about how the drift distance and drift sign is assigned. I was wondering if we wanted to separate the tracklet building process from the tracking part or this is just a change for the cosmic muon reconstruction only so that we can calibrate the X-T curve.

knakano0524:
@abinashpun Thank you for taking a look.  It is not to modify the default tracking algorithm (i.e. `SQReco`) but to create a special tracking mode (`SQTrackletReco`).  In the default tracking algorithm, the drift sign should not be resolved for each local tracklet, because it slows down the reconstruction.

I modified some lines of `SQReco` in order to only make it inheritable.  Its algorithm hasn't changed.

abinashpun:
Dear Kenichi,
Thank you. Then I think it is a good idea. Please let me know if there is anything you want me to look closer.

knakano0524:
I made another commit to clean up the code.  The code should be good enough to be merged.  @abinashpun and @liuk Could you take a look by the end of Friday?

I packed some common lines into member functions.  For example, `SQReco::ProcessEventPrep()` contains the preparation in `process_event()` and is called by `SQTrackletReco` as well as `SQReco`.

I modified `KalmanFastTrackletting::buildTrackletsInStation()`.  Now it does not use `acceptTracklet()` and not create duplicated tracklets.

abinashpun:
Dear Kenichi,
Thank you for the changes. I had look at the changes, did sanity check by building and running (CalibChambXT). Everything looks fine for me unless if you want me do any specific test.

knakano0524:
@abinashpun Thank you for making the sanity check.  That's helpful.

@liuk If you have any comment or need a time, please let me know by the end of today.  After that I will merge this PR.

PR #132, opened 2022-05-18
Speedup of Main-DAQ decoder and online monitor

knakano0524:
We found in the DAQ stress test in March that the decoder became slower than the spill cycle (1 m) under high rate.  This update is to identify time-consuming parts and improve the computation time.  The updated code is fast enough, I think.  I will post a presentation about this.  We can first discuss it at a commissioning(?) meeting and then you decide to merge this PR.

There are quite many lines modified, but most of them are simple.

In the first half of the modified lines (down to `online/macros/Fun4MainDaq.C`), variables and functions were changed/added to use a new timer, `PHTimer2`, for many code blocks.  The existing timer, `PHTimer`, was found to have two problems, as explained in `PHTimer2.h`.  In a future we should replace all of `PHTimer` with `PHTimer2`.

In the second half, I introduced a new helper function, `UtilSQHit::FindHitsFast()`, and modified all OnlMon clients (`online/onlmonserver/OnlMon*.cc`) to use the function.  It has an internal static variable that holds one hit vector per detector.  The variable is filled only on the 1st call per event, and used by all OnlMon clients.  This scheme reduces the computation time by 5x.  I haven't modified `SQHit` nor `SQHitVector` because I want to keep them simple as the fundamental data container.

At the very end, I modified the channel-mapping classes (`ChanMapTaiwan.h`, `ChanMapV1495.h` and `RunParamBase.h`) in order to speed them up.  `std::map` was replaced with `std::unordered_map`, which requires the hash function (`RocBoardChanHash`) but is indeed faster by 7x.

The output time can be reduced by 2x, by simply disabling the all-in-one DST (e.g. `run_003712_spin.root`).  It has not been disabled in the updated code, because I should first confirm that no one will need it.


knakano0524:
I posted the presentation:
https://seaquest-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=10005

For quick reference, attached is the updated OnlMonMainDaq plot;
OnlMonMainDaq_can2_3712
It shows the five computation times versus spill.

liuk:
Thanks Kenichi. While we are looking at the code, would you mind present the update next Monday?

liuk (approved):

PR #134, opened 2023-03-28
Addition of target materials

knakano0524:
This update is to add materials to the geometry, i.e. one target cup, three NMR coils and one ladder.  Zulkaida made the modifications, so he could give some more explanations.  He and Jay have compiled and executed the modified code fine.

### DIFF
--- a/macros/G4_Target.C
+++ b/macros/G4_Target.C
@@ -64,6 +64,87 @@ void SetupTarget(
   target->set_int_param("use_g4steps", use_g4steps);
   target->SetActive(register_hits);
   g4Reco->registerSubsystem(target);
+
+  PHG4CylinderSubsystem* target_cup = new PHG4CylinderSubsystem("Target", 1);
+  target_cup->SuperDetector("Target");
+  target_cup->set_double_param("length", target_l);
+  target_cup->set_double_param("rot_x", 0.);
+  target_cup->set_double_param("rot_y", 0.);
+  target_cup->set_double_param("rot_z", 0.);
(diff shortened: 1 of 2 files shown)


zulkaidaakbar:
We tried (compiled and Run) at UVA and it works

zulkaidaakbar:
Anyway, I don't know how to "approve" the review/code

knakano0524:
@zulkaidaakbar Please tell Abi and Kun roughly when you want to merge this PR, so that they can make a review in time.

I have slightly modified the code based on our e-mail conversation.  So please take a look at the "Files changed" and add comments/reviews if you want to modify further.

zulkaidaakbar:
@liuk @knakano0524 @abinashpun  No rush, sometime next week is okay for merging this request. This add 3 materials on the target: NMR coils, Target Cup, and Aluminum ladder which hold the cup

liuk:
@zulkaidaakbar have you run this macro and confirmed it will generate the geometry we are looking for? It seems to me that the empty target cup with NMR coils is generated correctly, but they are all parallel volumes in addition to the existing target material (which is also a cylinder). These two volumes should be in conflict with each other. Could you check it and show some vertex simulation plot to show that we can see all the target structures we are trying to simulate here?

zulkaidaakbar:
Target_Geometry
Hi @liuk Here I attach the geometry of the target. We can see clearly from the picture the NMR Coils, target cups (kelF), and the aluminum ladder. Could you please explain more about the vertex simulation to validate the geometry?

PR #135, opened 2023-04-19
V1495 Online Monitoring 

hazeltet845:
This is the V1495 online monitoring work that I showed in a presentation the other day. I removed the plots we talked. Please let me know if there is any more information you need from me.


knakano0524:
I'm so sorry for this late response.  I forgot to follow this up.  I will take a closer look this week.  But can you first clean up extra files like `bkp/` and maybe `rs/` also?

knakano0524:
I apologize for this very late follow-up.  I tested the code using run 5008 after a few modifications.  The outputs are attached.  Are they reasonable? 

you can find the modifications that I made here;
https://github.com/E1039-Collaboration/e1039-core/commit/31bc575c3283b3a08948cb729dbd898bc03f0bd8
One important change is that the roadset files were moved to the e1039-resource directory.

OnlMonTrigV1495_can0
OnlMonTrigV1495_can1
OnlMonTrigV1495_can2
OnlMonTrigV1495_can3
OnlMonTrigEP_can0

PR #136, opened 2023-12-12
Correct KMag position and better decoding error handling

knakano0524:
This PR includes two updates.

1. 
The KMag z-position was found incorrect, which should have been taken from an old E906 code.  It was corrected (from 1064.26 to 1042.01722 cm) in `recoConsts.cc`.  Also `PHFieldSeaQuest.cc` was modified not to use the hard-coded value in  it.

I added to `recoConsts::init(runNo)` the default parameter set for the E906 Run-6 data.  I plan to use this way of integrating all parameter sets for E1039 datasets.

2.
The Main-DAQ decoder was modified so that it can skip words from ROC or event in case of word-format errors, instead of discarding the whole event.  The modified code has already been used for the online decoding.

### DIFF
--- a/framework/phool/recoConsts.cc
+++ b/framework/phool/recoConsts.cc
@@ -146,7 +146,7 @@ void recoConsts::set_defaults()
   set_DoubleFlag("PT_KICK_KMAG", 0.4016);
   set_DoubleFlag("PT_KICK_FMAG", 2.909);
 
-  set_DoubleFlag("Z_KMAG_BEND", 1064.26);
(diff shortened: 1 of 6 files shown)


knakano0524:
Note:  The z-position of KMag defined at line 110 of `G4_InsensitiveVolumes.C` was/is correct (1042.01722).  But the position for the magnetic field was set at line 27 of `PHFieldSeaQuest.cc`.  Thus the field position in event generation and reconstruction were both wrong but consistent.  In addition the field strengths of FMag and KMag were the ones for the wrong position, thus the effect was compensated like the early E906 setting.

PR #138, opened 2024-02-16
Manual selection of plane geometry version.

knakano0524:
This update is mainly to add a method of manually selecting the plane geometry version.  Usually the plane geometry is selected based on the `RUNNUMBER` flag.  But if the `GeomPlane` flag exists (before instantiating `GeomSvc`) the ID given by `GeomPlane` is selected.


knakano0524:
I have confirmed that the modified code can be built without error, and the default behavior of the program hasn't changed.  Thus I merge this PR now.

PR #139, opened 2024-03-12
Inclusion of setup scripts.

knakano0524:
This update is just to add two scripts that are being used to set up the E1039 software (including "resource" and "share") on E1039 servers.  The scripts are not in any repository at present.


knakano0524:
This update won't change the behavior of the core program.  So I merge it by myself.

PR #140, opened 2024-05-02
Change in the condition when OnlMonServer suicides

knakano0524:
This update is to fix a problem that the OnlMonServer sometimes becomes unavailable.  OnlMonServer doesn't suicide the server process if its run number is larger than the one that sends the suicide request.


knakano0524:
The updated version is already running for the online decoding without problem.

PR #145, opened 2024-08-13
Transition from Scientific Linux 7 (SL7) to AlmaLinux 9 (AL9)

knakano0524:
This update is to switch the default OS from SL7 to AL9.  It cannot be automatically merged, so I will update and test the merged coder later.


knakano0524:
I tested the updated code on `e1039gat1` and `spinquestgpvm01`.  The event generation and analysis (`SimChainDev` and `AnaSimDst`) worked fine.  The event generation ran on GRID.

Note that the conflict was quite simple and fixed.  I had made a change in `CalibHitElementPos.cc` on both "master" and "al9" but then modified only "master".

PR #148, opened 2024-10-05
Control option for hit-level hodo masking in KalmanFastTracking

knakano0524:
This update is mainly to add a constant to `recoConsts` to control the hit-level hodo masking in `KalmanFastTracking`.

The new constant, `HIT_MASK_MODE`, is `AUTO` by default, which keeps the original behavior of the hit-level hodo masking, namely X-hodo hits are used when `MC_MODE` is true, `COSMIC_MODE` is true or event is triggered by FPGA.

When it is set to `X`, `Y` or `XY`, the specified hodo plane is always used.  When it is set to `NONE`, the hit-level hodo masking is disabled.

A new (4th) argument was added to the constructor of `KalmanFastTracking` in order to change the verbosity in the constructor.

The default value of `SRawEvent::fTriggerBits` was changed from `-1` to `0`, because `-1` means all bits are `1` by default, which seemed strange.  Indeed it had caused a problem as `SRawEvent::SetTriggerBits()` hadn't changed bit from `1` to `0`.

### DIFF
--- a/framework/phool/recoConsts.cc
+++ b/framework/phool/recoConsts.cc
@@ -53,6 +53,7 @@ void recoConsts::set_defaults()
   set_BoolFlag("MC_MODE", false);
   set_BoolFlag("COSMIC_MODE", false);
   set_BoolFlag("REQUIRE_MUID", true);
+  set_CharFlag("HIT_MASK_MODE", "AUTO");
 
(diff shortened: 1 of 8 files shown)


liuk:
Looks good Kenichi, please go ahead with the merge.

PR #149, opened 2024-10-13
New SQVertexing package to (hopefully) handle dimuon vertexing correctly

liuk:
I created a new package in ```packages/reco/ktracker``` named ```SQVertexing``` to handle the vertex finding and dimuon vertex fitting. 

# General usage
The primary usage is like this:

```
SQVertexing* vertex = new SQVertexing();
//Choose to save the output dimuons inside SRecEvent (if true), or in a Fun4All container (if false), default is false
vertex->set_legacy_rec_container(false); 
//Choose to update the target swim results with the ones after applying track refitting, default is false
vertex->set_single_retracking(false);
//Specify the input file name for ROOT geometry, only needed if it's run without track reconstruction, i.e. run without SQReco in the module list
vertex->set_geom_file_name("filename_for_ROOT_geometry.root")
se->registerSubsystem(vertex);
```

# Primary change and results
Under the hood, the primary change compared with the E906 legacy ```VertexFit``` is it utilizes single muon bend plan correction instead of the old "dimuon mass optimization" process. The single muon bend plane in FMag is decided entirely based on MC truth, based on the momentum of the muon. 

Here is some comparison of the mass resolution, single track $`P_x/P_z`$ resolution, and vertex resolution. All plots are based on DY MC from 3-9 GeV.

Snipaste_2024-10-12_06-56-29
The plot above shows the relative mass resolution, which is defined as $`(M_{rec} - M_{true})/M_{true}`$.  Black line shows the result if both muons are constrained to go to the reconstructed `Zvtx`. Blue line shows the result if both muons are forced to go to the target center without bend plane correction. Red line shows the result if both muons are forced to go to the target center with the bend plane correction.

Snipaste_2024-10-12_07-50-49
Similarly, we have the single muon $`P_x/P_z`$ resolution shown in the plot above. The definition of the colors is the same.

Snipaste_2024-10-13_05-59-06
We also have the reconstructed vertex resolution (~45cm) shown in the plot above.

# Changes to the data structure
In order to preserve all the information and leave them for the analyzer to decide. I made a change to the ```SRecDimuon``` data structure.  Four new ```TLorentzVector```s are added, combined with the original two we now have 6 to store the dimuon track information under different z_vertex hypothesis. At this the vertex fit package performs three hypothesis tests:
- Choice 1: assuming both tracks come from reconstructed $`Z_{vtx}`$, the result is saved in ```p_pos``` and ```p_neg```;
- Choice 2: assuming both tracks come from target center, the result is saved in ```p_pos_target``` and ```p_neg_target```;
- Choice 3: assuming both tracks come from dump face, the result is saved in ```p_pos_dump``` and ```p_neg_dump```;

Afterwards the use can access reconstructed kinematic variables ($`mass, x_F`$, etc.) by calling ```SRecDimuon::calcVariables(choice)```. The choice number in the argument refers to the choices listed above. Without any arguments the variables are calculated using momentum obtained by fitting both tracks to reconstructed $`Z_{vtx}`$. 

# To-do
This is the first attempt to get the vertex and dimuon reconstruction correct. One can see that even the muon angles are reconstructed very well, there is still a slight (~50MeV) shift in the reconstructed mass. This is believed to be due to the incorrect  dE/dx correction. This correct was performed by Abi 3-4 years ago but needs to be double-checked and also have the concrete shielding blocks included as well.

Also this package does not handle the need for like-sign dimuon reconstruction, which needs to be included in the future.

### DIFF
--- a/framework/phool/recoConsts.cc
+++ b/framework/phool/recoConsts.cc
@@ -163,7 +163,7 @@ void recoConsts::set_defaults()
   set_DoubleFlag("FMAG_HOLE_LENGTH", 27.94);
   set_DoubleFlag("FMAG_HOLE_RADIUS", 1.27);
   set_DoubleFlag("FMAG_LENGTH", 502.92);
-  set_DoubleFlag("Z_UPSTREAM", -500.);
+  set_DoubleFlag("Z_UPSTREAM", -700.);
   set_DoubleFlag("Z_DOWNSTREAM", 500.);
 }
 
--- a/packages/reco/interface/SRecEvent.cxx
+++ b/packages/reco/interface/SRecEvent.cxx
@@ -608,6 +608,23 @@ void SRecDimuon::calcVariables(int choice)
         pos = p_pos_dump;
         neg = p_neg_dump;
     }
+    else
+    {
+        std::cout << "ERROR!!! In SRecDimuon::calcVariables(choice), only three dimuon vertex choice numbers are provided - ";
+        std::cout << " 0 (default) = fitted vertex position, 1 = target position, 2 = dump position" << std::endl;
+        
+        //Unphysical error values
+        mass = -1.;
+        xF = -10.;
+        x1 = -10.;
+        x2 = -10.;
+        pT = -1.;
+        costh = -10.;
+        phi = -10.;
+        mass_single = -1.;
+
+        return;
+    }
 
     Double_t mp = 0.938;
     Double_t ebeam = 120.;
--- a/packages/reco/interface/SRecEvent.cxx
+++ b/packages/reco/interface/SRecEvent.cxx
@@ -590,8 +590,25 @@ void SRecTrack::print(std::ostream& os) const
   os << "Chi square at vertex: " << fChisqVertex << std::endl;
 }
 
-void SRecDimuon::calcVariables()
+void SRecDimuon::calcVariables(int choice)
 {
+    TLorentzVector pos, neg;
+    if(choice == 0)
+    {
+        pos = p_pos;
+        neg = p_neg;
+    }
+    else if(choice == 1)
+    {
+        pos = p_pos_target;
+        neg = p_neg_target;
+    }
+    else if(choice == 2)
+    {
+        pos = p_pos_dump;
+        neg = p_neg_dump;
+    }
+
     Double_t mp = 0.938;
     Double_t ebeam = 120.;
 
@@ -599,7 +616,7 @@ void SRecDimuon::calcVariables()
     TLorentzVector p_target(0., 0., 0., mp);
 
     TLorentzVector p_cms = p_beam + p_target;
-    TLorentzVector p_sum = p_pos + p_neg;
+    TLorentzVector p_sum = pos + neg;
 
     mass = p_sum.M();
     pT = p_sum.Perp();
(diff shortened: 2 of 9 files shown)


framework/phool/recoConsts.cc
```diff
@@ -163,7 +163,7 @@ void recoConsts::set_defaults()
   set_DoubleFlag("FMAG_HOLE_LENGTH", 27.94);
   set_DoubleFlag("FMAG_HOLE_RADIUS", 1.27);
   set_DoubleFlag("FMAG_LENGTH", 502.92);
-  set_DoubleFlag("Z_UPSTREAM", -500.);
+  set_DoubleFlag("Z_UPSTREAM", -700.);
```
knakano0524, line 166: How much does this change affect the target-upstream separation (i.e. chi2_target > or < chi2_upstream)?

packages/reco/interface/SRecEvent.cxx
```diff
@@ -590,16 +590,33 @@ void SRecTrack::print(std::ostream& os) const
...
+    else if(choice == 1)
+    {
+        pos = p_pos_target;
+        neg = p_neg_target;
+    }
+    else if(choice == 2)
+    {
+        pos = p_pos_dump;
+        neg = p_neg_dump;
+    }
```
knakano0524, line 610: Can you add an "else" case to show a warning (or an error to abort) if `choice` is invalid?

packages/reco/interface/SRecEvent.h
```diff
@@ -343,8 +343,8 @@ class SRecDimuon: public SQDimuon
     //Get the total momentum of the virtual photon
     TLorentzVector getVPhoton() { return p_pos + p_neg; }
 
-    //Calculate the kinematic vairables
-    void calcVariables();
+    //Calculate the kinematic vairables, 0 = vertex, 1 = target, 2 = dump
```
knakano0524, line 346: Can you replace `//` with `///` so that the comment gets visible on the Doxygen page?

knakano0524 (commented):
I have a few minor comments.  I want to check whether/how the change of `Z_UPSTREAM` affects the default/past reco. result (using `VertexFit`).  It it is not significant, we can merge this PR right now, so that many people can test `SQVertexing`.  If it is, is the new value (-700) the best or at least better?

framework/phool/recoConsts.cc
liuk: Actually yet, moving the Z_UPSTREAM by 200cm improves the target/upstream separation. See the plot below.

Snipaste_2024-10-15_02-30-48
The blue line is $`\chi^{2}_{upstream}-\chi^{2}_{target}`$ when upstream is set to -500 cm, the red line is the same distribution when upstream is moved to -700 cm.

packages/reco/interface/SRecEvent.cxx
liuk: Good advice, added another commit

packages/reco/interface/SRecEvent.h
liuk: Good advice, changed the comment style in another commit

framework/phool/recoConsts.cc
knakano0524: While the change improves the efficiency of target/upstream separation, it worsens the purity simultaneously, right?  I remember the beam window is around -500 cm.  The guessed number is fine for now, but we will need to know the locations of upstream objects, and probably run the GMC study.

knakano0524:
Thank you for the answers and the updates.  Can you merge this PR?

PR #150, opened 2024-10-28
Updates on SQVertexing, Fun4AllSRawEventOutputManager and UtilSRawEvent

knakano0524:
This pull request includes multiple updates eventually.

1. SQVertexing was found not converged on some events of the commissioning data.  To avoid it, we added lines to `SQVertexing::calcZsclp()` to enforce the valid momentum range, and also added the max number of iterations to `SQVertexing::refitTrkToVtx()`.  Verbose messages were added for debugging.

2. `Fun4AllSRawEventOutputManager` was renamed as `Fun4AllSpillSRawEventOutputManager` because it outputs spill-by-spill files.  A new manager, `Fun4AllSRawEventOutputManager`, was made, which outputs one file per process (=run).

3. A bug about the trigger bits in `UtilSRawEvent::SetEvent()` was fixed.  The order of FPGA and NIM bits in `SRawEvent` is different from that in `SQEvent`, but the function above had assumed that it is identical.

### DIFF
--- a/online/macros/Fun4MainDaq.C
+++ b/online/macros/Fun4MainDaq.C
@@ -116,7 +116,7 @@ int Fun4MainDaq(const int run=46, const int nevent=0, const bool is_online=false
   }
 
   if (is_online) {
-    Fun4AllSRawEventOutputManager *om_sraw = new Fun4AllSRawEventOutputManager("/data4/e1039_data/online");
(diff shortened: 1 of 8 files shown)


liuk:
Thanks Kenichi, I just made some very minor format tweak and also adjusted the upper limit of Z_VTX from 231cm to 331cm. Please go ahead with this PR.

PR #151, opened 2024-11-12
Add flexibility in SQVertexing IO options

liuk:
This PR made two changes:

1. It added additional Boolean flags (and associated set functions) to allow user to control whether to use legacy-style input/output containers separately. ```SQVertexing::set_legacy_in_container()``` will switch between using ```SQTrackVector``` and old ```std::vector<SRecTrack>``` embedded inside the ```SRecEvent``` as input. Similarly,  ```SQVertexing::set_legacy_out_container()``` will switch between using ```SQDimuonVector``` and old ```std::vector<SRecDimuon>```  embedded inside the ```SRecEvent``` as output. Setting ```SQVertexing::set_legacy_container()``` will turn on/off both input and output switch simultaneously. 
2. Added some code at the end of ```SQVertexing::processOneDimuon()``` to change the sign of the Px of the "positive" muon if we are processing like-sign dimuons.

### DIFF
--- a/packages/reco/ktracker/SQVertexing.cxx
+++ b/packages/reco/ktracker/SQVertexing.cxx
@@ -58,7 +58,6 @@ namespace
 
 SQVertexing::SQVertexing(const std::string& name, int sign1, int sign2):
   SubsysReco(name),
-  legacyContainer(false),
   legacyContainer_in(false),
   legacyContainer_out(false),
   enableSingleRetracking(false),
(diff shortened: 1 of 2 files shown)


knakano0524:
Thank you for making the necessary changes quickly.  I just wonder if we can remove `bool legacyContainer` from SQVertexing.

liuk:
> Thank you for making the necessary changes quickly. I just wonder if we can remove `bool legacyContainer` from SQVertexing.

Thanks. We certainly can, it was kept to keep the current code working. But since you are probably the only one using this new package, we can certainly remove it. Let me know if you prefer it this way.

liuk:
> I think we can keep only the setter function like `void set_legacy_rec_container(const bool enable = true) { legacyContainer_in = enable; legacyContainer_out = enable; }`. Then it should be backward compatible.

Of course, this makes much better sense. I have just pushed another commit.

PR #153, opened 2024-11-25
Initialization of GFTrack

knakano0524:
This is to fix a problem that SQVertexing outputs a different/strange vertexing result when it is used alone (i.e. without SQReco).  One clear failure was seen on the matrix inversion made in "GFTrack::updatePropState()".  It was because the beam covariance matrix was zero, and the root cause was that GFTrack was initialized only in the default constructor.  The default constructor is called only in SQReco, thus this problem was not apparent.


knakano0524:
I added the following lines also.
```
    genfit::FieldManager::getInstance()->init(gfield);
    genfit::MaterialEffects::getInstance()->init(new genfit::TGeoMaterialInterface());
```

PR #154, opened 2024-12-05
Fixed the bias in always selecting mu+ for Px sign flip in like-sign dimuon processing

liuk:
We now use a random generator (TRandom2) to select which muon we choose to flip the sign of Px when processing like-sign dimuons, because apparently the track list is pre-sorted by track quality (reduced chi2) thus creating a bias.


knakano0524:
Thank you for the quick update.  It looks good to be merged.

PR #156, opened 2024-12-17
Single Muon Generator

liuk:
A quick version of a single muon generator. The pion momentum and pT distributions are digitized from this paper https://arxiv.org/pdf/1404.5882. The pion is sequentially decayed to muons. A new Truth Information (SQSingleMuonTruthInfo) node is added to contain the generator level information.


knakano0524:
The code looks fine to be merged.  It would be better in future if we put all generator classes in one directory.

PR #160, opened 2025-03-07
EMCal bug fix

yongbinfeng:
PR to fix two issues in the current EMCal simulation. 

- First one is on the EMCal geometry, since the unit is already specified in the constructor, adding another unit would cause the geometry to be wrong by one order of magnitude.
- Second is on the energy deposit of EMCal hit. Without setting to zero, it would be randomly initialized, causing some undefined behaviors.

The fixes have been tested multiple times for the EMCal simulations for dark sector. Made this PR for synchronization with the master branch. PR should not affect anything except EMCal.


yongbinfeng:
@knakano0524 ping, can this be merged? should not affect anything on current spinquest.

knakano0524:
Thank you for the reminder.  The changes look fine.  I have merged it.

PR #161, opened 2025-05-19
 adding true track variables in the RUS output file

forhadnmsu:
## True Track Variable Output in Fun4AllRUSOutputManager

This version adds variables from the true track information using the `SQTrack` class node. In addition to including the `trackID` in the true track label, the `trackID` is also added to the hit label. The name of the variable is `hit_trackID`.

In a Fun4All macro, the output manager should work for both simulation and experimental data. For this version, the `processID` is set manually. I will update this in the next version by setting those variables automatically by parsing the SQ nodes inside `Fun4AllRUSOutputManager`.

### Process ID Values
- DY = 11 (mu⁺)
- J/psi = 12 (mu⁺)
- psi′ = 13 (mu⁺)
- Single muon = 14

For the corresponding mu⁻ particle, the `processID` is set as +10 the above value.

The process ID in the hit label is encoded along with the source flag using the following logic:

```cpp
// Encoding
unsigned int Fun4AllRUSOutputManager::EncodeProcess(int processID, int sourceFlag) {
    return ((sourceFlag & 0x3) << 5) | (processID & 0x1F);
}

//To decode the encoded process ID and source flag, use the following functions:

```cpp
int DimuAnaRUS::DecodeSourceFlag(unsigned int encoded) {
    return (encoded >> 5) & 0x3; // Extracts the 2-bit source flag
}

int DimuAnaRUS::DecodeProcessID(unsigned int encoded) {
    return encoded & 0x1F; // Extracts the 5-bit process ID
}

//Example Usage
//Below is an example of how to configure and register the `Fun4AllRUSOutputManager` in a Fun4All macro:

```cpp
Fun4AllRUSOutputManager* rus_out = new Fun4AllRUSOutputManager();
rus_out->SetTreeName("tree");
rus_out->SetProcessId(14); // Use: DY = 11, J/psi = 12, psi' = 13, single muon = 14
rus_out->SetMCTrueTrackMode(true); // Set to false to omit true track variables
rus_out->DisableSQEvent(true); // Set to false for experimental data
rus_out->SetFileName("RUS.root");
se->registerOutputManager(rus_out);

### DIFF
--- a/packages/RUS/Fun4AllRUSOutputManager.cc
+++ b/packages/RUS/Fun4AllRUSOutputManager.cc
@@ -88,7 +88,7 @@ int Fun4AllRUSOutputManager::OpenFile(PHCompositeNode* startNode) {
 
 
     if (true_mode) {
-        m_tree->Branch("gprocessID", &gprocessID); //Hit level
+        m_tree->Branch("gProcessID", &gProcessID); //Hit level
         m_tree->Branch("gCharge", &gCharge);
         m_tree->Branch("gTrackID", &gTrackID);
         m_tree->Branch("gvx", &gvx);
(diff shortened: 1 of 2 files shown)


knakano0524:
The code looks fine overall.  Have you built the modified code for confirmation?

Can you check if the names of the new variables are fine in terms of the naming convention?  As an example the followings seem to be more suitable;
* `trackID` -> `gTrackID`
* `processID` -> `gProcessID`
* `hit_trackID` -> `hitTrackID`

PR #162, opened 2025-05-19
Small updates on trigger utility and legacy generator

knakano0524:
This update is to make multiple small changes.  I have confirmed that the updated code can be built fine and `SQPrimaryParticleGen` sets the process ID of `SQMCEvent` properly.

I added functions to `UtilTrigger`, which should be useful in many analyses.  One big function is to extract a list of all fired enabled roads from all (or only in-time) hodoscope hits.

I gave IDs (codes) to the processes generated by `SQPrimaryParticleGen`.  The IDs are defined as `ProcessCode_t`, where they are negative not to overlap with the codes used in Pythia8.  The ID is accessible in analysis via `SQMCEvent::get_process_id()`.  Inside `SQPrimaryParticleGen` the generated process is selected via `_proc_code` instead of `_PythiaGen` etc. and thus only one process can be enabled now.

I replaced `SQEvent_v1` with `SQEvent_v2` in simulation-related classes.  `SQEvent_v2` is slightly smaller since it doesn't contain real-data-only variables.

I fixed typos in debug messages of `SQVertexing`.

### DIFF
--- a/generators/E906LegacyGen/SQPileupGen.C
+++ b/generators/E906LegacyGen/SQPileupGen.C
@@ -26,7 +26,7 @@
 #include <g4main/PHG4TruthInfoContainer.h>
 #include <Geant4/G4ParticleTable.hh>
 #include <interface_main/SQMCEvent_v1.h>
-#include <interface_main/SQEvent_v1.h>
+#include <interface_main/SQEvent_v2.h>
 
 #include "SQPileupGen.h"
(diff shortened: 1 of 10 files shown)


liuk:
Thanks Kenichi. All look good to me.

forhadnmsu:
All process IDs for PYTHIA8 are positive, so the convention looks good to me for setting the process ID.

knakano0524:
@liuk @forhadnmsu Thank you for the prompt checks.

PR #163, opened 2025-05-19
Small changes in Single Muon generator, added single muon vertexing

liuk:
This PR consists of three changes. 

The first commit changes the single muon generator to read X_BEAM and Y_BEAM reco consts when generating the primary vertex. In the past the primary vertex X and Y positions are hard-coded to be 0.

The second and third commit moves the single muon vertexing code from SQGenFit::GFTrack::getSRecTrack to SQVertexing::processOneMuon. In the past the single muon vertexing is done in SQGenFit::GFTrack::getSRecTrack which is called by SQReco. After we update the X_BEAM and Y_BEAM information and re-run the production with only SQVertexing the single muon info will not be updated.

The last commit added an error count data member is SQGenFit::GFTrack to supress the excessive error message in the extrapolation. Now the code will suppress future error messages after 5 errors in the same track.

### DIFF
--- a/generators/SingleMuonGen/SQSingleMuonGen.cxx
+++ b/generators/SingleMuonGen/SQSingleMuonGen.cxx
@@ -11,6 +11,7 @@
 #include <phool/PHCompositeNode.h>
 #include <phool/PHIODataNode.h>
 #include <phool/PHRandomSeed.h>
+#include <phool/recoConsts.h>
 
 #include <Geant4/G4ParticleTable.hh>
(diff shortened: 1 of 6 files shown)


knakano0524:
All the changes look fine and useful.  Could you just delete the old code (i.e. lines 442-480 of `GFTrack.cxx`) now?  Or decide when we can delete them?

I will anyway merge this PR on tomorrow morning, and build a new offline version including PR #161-163.

PR #167, opened 2025-11-07
Non-Fun4All use of SQVertexing and Function of road matching with margin

knakano0524:
1. `SQVertexing` was modified so that it can be used outside the Fun4All macro.  It is necessary in the track mixing (i.e. `e1039-analysis/TrackMixingNMSU`) because events are mixed (sorted) before vertexing.  The present analysis code carries a modified version of SQVertexing (`_v2`).  We will no longer need it and maintain this original version only.
2. `UtilTrack::FindMatchedRoads()` was added.  It finds all roads that match to the given track within the hodo paddle width plus a given margin.  The road ID currently assigned to `SRecTrack` is single with no margin considered.
3. Interface classes under `interface_main` were modified to suppress warning (i.e. member functions were returning local objects).

The modified code has been complied fine.  The default behavior of all programs is unchanged.  The performance of `FindMatchedRoads()` will be studied.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -64,8 +64,6 @@ else # 'all' or 'resume'
     framework/ffaobjects
     framework/fun4all
     interface_main
-    packages/UtilAna
-    online/decoder_maindaq
     packages/Half
     packages/vararray
     database/pdbcal/base
@@ -79,8 +77,6 @@ else # 'all' or 'resume'
     simulation/g4main
     simulation/g4detectors
     simulation/g4eval
-    generators/E906LegacyGen
-    generators/SingleMuonGen
     packages/calibrator
     packages/evt_filter
     packages/dptrigger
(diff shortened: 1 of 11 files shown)


forhad-h2030:
I had a look at `SQVertexing.cc`, and it looks good to me.  Since `e1039-analysis/TrackMixingNMSU` is designed to run without DST dependencies,  have you tested a sample mixing run with it to confirm that the modified `SQVertexing` works smoothly?

knakano0524:
Thank you for the comment.  Actually I hadn't tested the new version.  Now I found and fixed a problem (calling private member functions), and confirmed using run 6111 that the new version results in the same unmixed and mixed mass distributions as the original version.

forhad-h2030:
> Thank you for the comment. Actually I hadn't tested the new version. Now I found and fixed a problem (calling private member functions), and confirmed using run 6111 that the new version results in the same unmixed and mixed mass distributions as the original version.

Thanks for the fix. I’ve used your updated pull request version to run the `SQVertexing` for two DY simulated target events. I also saved the RUS-format hit information for cross-check.
Then, I used the `Fun4AllRUSInputManager` to load the RUS data and ran the `SQVertexing` with the current Git version of `SQVertexing`. The reconstructed masses are identical in both cases.  To me, it's fine to merge with the master branch.

knakano0524:
Thank you for the test.  It is helpful for confirmation.  I will marge this PR now.

PR #170, opened 2026-07-06
NRQCD J/psi Generator

knakano0524:
This update is to add a J/psi generator that generates J/psi following $d\sigma/dx_F$ predicted by NRQCD.

The generator doesn't compute NRQCD itself, but reads the cross-section tables under e1039-resource/generator/Jpsi_NRQCD.  The tables were provided by Wen-Chen Chang.  One key feature is that the cross sections in p+p, p+d and p+N are available.

The new code has been built fine, and doesn't affect any existing codes.


liuk:
Thanks Kenichi, looks all reasonable to me.

